PR #1, opened 2018-01-09
changed styled to glamorous & added components

marksy:
- added layout component
- styled checkbox


package.json
```diff
@@ -2,6 +2,7 @@
   "name": "govuk-react",
   "version": "0.1.3",
   "dependencies": {
+    "glamorous": "^4.11.2",
```
penx: please also remove styled-components

src/components/layout/index.js
```diff
@@ -0,0 +1,60 @@
+// https://github.com/alphagov/govuk-frontend/blob/master/src/components/checkbox/_checkbox.scss
```
penx: this first line should be a link to the scss file (or other related component resources) that you are basing this component on

src/components/checkbox/index.js
```diff
@@ -1,21 +1,79 @@
...
-const WrapperLabel = styled.label`
-  display: block;
-`;
 
 const Checkbox = ({children, ...props}) => (
-  <WrapperLabel>
-    <Input type="checkbox" {...props} />
-    {children}
-  </WrapperLabel>
+  <MultipleChoice>
```
penx: Why change this so it's no longer wrapped with a label?

Having a non wrapping label means that props.id becomes a required prop, which would need to be represented by propTypes, but personally I think is unnecessary if we can use a wrapping label instead

marksy: i was following what is on the gov.uk repo, so it's easier to grab their styles. Having an id on each input makes it easier to grab the value no?

penx:
As discussed

- we need to think about how we will handle breakpoints but can address this later
- in future PRs we should try to do 1 component at a time and have unit tests that prove that each prop the component can take has an effect on the output
- we should favour props rather than undocumented structure of the `children` dom content (e.g. for `<HintText />`) but can address this later
- we should start using proptypes and ensure all components props are included but can do this later

PR #9, opened 2018-01-17
DRYing out some of the styling

marksy:
- moved the some of the styling into constants file so it can be reused across other components
- created new smaller components; label, labelText, hintText, errorText, input
- refactored the components with reusable smaller components

### DIFF
--- a/src/components/fileUpload/index.js
+++ b/src/components/fileUpload/index.js
@@ -5,104 +5,21 @@ import React from "react";
 import glamorous from "glamorous";
 import PropTypes from "prop-types";
 
-import * as COLOUR from "govuk-colours";
-import { BREAKPOINTS, SITE_WIDTH } from "../../constants/index";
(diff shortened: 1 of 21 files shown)


src/components/fileUpload/index.js
```diff
@@ -5,104 +5,21 @@ import React from "react";
...
-    paddingBottom: "2px",
-    [mediaQueries.largeScreen]: {
-      fontSize: "19px",
-      lineHeight: "1.31579"
-    }
-  },
-  ({ errorText }) => ({
-    fontWeight: errorText ? 700 : LabelText.fontWeight
-  })
+const LabelText = glamorous.span(STYLE_DEFAULT.LABEL_TEXT, ({ errorText }) =>
```
penx: this feels as though LabelText should be its own component in its own file rather than reconstructed with the same value in multiple places

marksy:
no longer a WIP

PR #12, opened 2018-01-18
added defaultChecked & disabled to Radio & Checkbox

marksy:
- fixed additional props (disabled & pre-selected)

Merge branch 'master' into feature/dryStyling

### DIFF
--- a/src/components/checkbox/index.js
+++ b/src/components/checkbox/index.js
@@ -82,11 +82,7 @@ const Label = glamorous.span({
 
 const Checkbox = ({ children, ...props }) => (
   <MultipleChoice>
-    <Input
-      type="checkbox"
-      checked={props.defaultChecked}
-      disabled={props.disabled}
-    />
+    <Input type="checkbox" {...props} />
     <Label>{children}</Label>
   </MultipleChoice>
 );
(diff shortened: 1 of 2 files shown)


src/components/checkbox/index.js
```diff
@@ -80,15 +80,22 @@ const Label = glamorous.span({
   }
 });
 
-const Checkbox = ({ children }) => (
+const Checkbox = ({ children, defaultChecked, disabled }) => (
   <MultipleChoice>
-    <Input type="checkbox" />
+    <Input type="checkbox" checked={defaultChecked} disabled={disabled} />
```
penx: where we are using inputs, we want to pass quite a few different props straight down to the Input component.

Rather than naming each one, we can instead name the props that we don't want to pass down (children) and then pass down ...rest.

So:

```
const Checkbox = ({ children, ...props }) => (
    <MultipleChoice>
     <Input type="checkbox" ...props} />
     <Label>{children}</Label>
    </MultipleChoice>
```

src/components/radio/index.js
```diff
@@ -87,16 +87,28 @@ const LabelText = glamorous.span({
   }
 });
 
-const Radio = ({ children, name }) => (
-  <Label>
-    <Input type="radio" name={name} />
+const Radio = ({ children, ...props }) => (
+  <Label {...props}>
+    <Input
```
penx: I think we should pass ...props to the input rather than the lave

PR #13, opened 2018-01-19
Feature/back link

marksy:
- added new back link component

### DIFF
--- a/src/components/backLink/index.js
+++ b/src/components/backLink/index.js
@@ -1,9 +1,5 @@
 // References:
-// https://govuk-elements.herokuapp.com/buttons/
-//
-// https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/design-patterns/_buttons.scss
-// https://github.com/alphagov/govuk-frontend/blob/master/src/components/button/_button.scss
-// https://github.com/alphagov/govuk_elements/blob/master/packages/govuk-elements-sass/public/sass/elements/_buttons.scss
+// https://github.com/alphagov/govuk-frontend/tree/master/src/components/back-link
 
 import React from "react";
 import PropTypes from "prop-types";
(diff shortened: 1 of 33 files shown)


src/components/backLink/index.js
```diff
@@ -0,0 +1,61 @@
+// References:
+// https://govuk-elements.herokuapp.com/buttons/
+//
+// https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/design-patterns/_buttons.scss
```
penx: is the back link derived from the button css or is this copy pasta?

src/components/backLink/stories.js
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import { action } from "@storybook/addon-actions";
+// import iconPointer from "../../../images/icon-pointer.png";
```
penx: remove commented out code, also avoid import from png if you can e.g. make svg components

src/components/fileUpload/index.js
```diff
@@ -5,104 +5,21 @@ import React from "react";
...
-  },
-  "[disabled]": {
-    cursor: "auto"
-  },
-  ":focus": {
-    outline: `3px solid ${COLOUR.YELLOW}`,
-    outlineOffset: 0
-  }
-});
+const Input = glamorous.input(STYLE_DEFAULT.INPUT);
```
penx: should this be a standard input component?

marksy: i've just merged master into this branch to get the latest which solves this issue

src/components/backLink/index.js
penx: could you check that similar links in the project are correct and not copy pasta too?

PR #17, opened 2018-01-22
new component panel

marksy:
- new panel component

basic header/panel component with mandatory heading and optional body that accepts html markup


penx:
Missing unit tests

package.json
```diff
@@ -4,9 +4,9 @@
   "dependencies": {
     "glamorous": "^4.11.2",
     "govuk-colours": "^1.0.2",
+    "jest-glamor-react": "^3.3.0",
```
penx: jest should be a devDependency

package.json
```diff
@@ -4,9 +4,9 @@
...
     "prop-types": "^15.6.0",
-    "react": "^15.0.0",
-    "react-dom": "^15.0.0"
+    "react": "^15.6.2"
```
penx: we eventually need to move react in to:

1. devDependencies
2. peerDependencies with a large range

..but this doesn't need to be done now

package.json
```diff
@@ -46,19 +50,21 @@
     "babel-preset-stage-1": "^6.24.1",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-15": "^1.0.5",
+    "enzyme-to-json": "^3.3.0",
     "eslint": "^4.15.0",
     "eslint-config-airbnb": "^16.1.0",
     "eslint-config-prettier": "^2.9.0",
     "eslint-plugin-import": "^2.8.0",
-    "eslint-plugin-jsx-a11y": "^6.0.3",
+    "eslint-plugin-jsx-a11y": "^5.1.1",
```
penx: why the downgrade?

src/components/backLink/test.js
```diff
@@ -5,8 +5,9 @@ import ReactDOM from "react-dom";
 import BackLink from "./";
 
 describe("Back Link", () => {
+  const testFn = () => false;
```
penx: this test should use sinon to confirm that the click handler is called when the button is clicked, but happy for this to go in as a TODO: comment for now

PR #18, opened 2018-01-22
CreateNewComponent script

marksy:
- created a node script to generate `MyNewComponent` folder, with an index.js (component), test.js (test) and stories.js (story)

- updated readme to explain how to generate a new component


src/components/createComponent.js
```diff
@@ -0,0 +1,96 @@
...
+const createFolder = () => {
+  mkdirp(folderName, err => {
+    if (err) {
+      console.error(err);
+    }
+  });
+};
+
+const writeThis = (fileName, contents) => {
+  fs.writeFile(`./${folderName}/${fileName}.js`, `${contents}`, err => {
```
penx: scripts like this will need to work in Windows so should use path.join rather than '/'

penx:
Please take a look at the this, give it a try, and consider whether it would be worth us using it instead of rolling our own:

https://github.com/CVarisco/create-component-app

.eslintrc
```diff
@@ -14,7 +14,7 @@
   },
   "overrides": [
     {
-      "files": [ "stories.js", "test.js", "src/stories/index.js", "src/setupTests.js" ],
+      "files": [ "stories.js", "test.js", "src/stories/index.js", "src/setupTests.js", "src/components/createComponent.js" ],
```
penx: build scripts, imo, should live in /scripts rather than /src, or potentially in /src/scripts/

There should also be an npm script that runs this, e.g. `npm run create-component`

scripts/createComponent.js
```diff
@@ -0,0 +1,116 @@
...
+  const tempFileName = `${fileName}.js`;
+  const pathName = path.join(folderName, tempFileName);
+
+  fs.writeFile(pathName, `${contents}`, err => {
+    if (err) {
+      return console.log(err);
+    }
+    return false;
+  });
+  console.log(`Created file: ${folderName}/${fileName}.js`);
```
penx: +  console.log(`Created file: ${pathName}`);

PR #39, opened 2018-01-29
Feature/expose component to parent

marksy:
- fixed padding on grid component
- fixed tests
- added props to input
- added props to textarea


src/components/Input/index.js
```diff
@@ -36,4 +37,18 @@ const Input = glamorous.input(
   })
 );
 
+const Input = ({ value, onChange }) => (
+  <InputText type="text" value={value} onChange={onChange} />
+);
+
+Input.defaultProps = {
+  value: undefined,
+  onChange: null
```
penx: maybe make this undefined rather than null as null in snapshots looks ugly

penx: in fact we can probably just export the InputText glamorous component and set a default prop of type to text, deleting the input component which is purely used as a wrapper other than type=text

src/components/InputField/index.js
```diff
@@ -10,28 +10,47 @@ import ErrorText from "../ErrorText/index";
...
-  <Label errorText={errorText}>
-    <LabelText errorText={errorText}>{children}</LabelText>
-    {hintText ? <HintText>{hintText}</HintText> : <span />}
-    {errorText ? (
-      <ErrorText errorText={errorText}>{errorText}</ErrorText>
-    ) : (
-      <span />
-    )}
-    <Input type="text" errorText={errorText} />
+const InputField = props => (
```
penx: I know I may have written this, but looking at it now I think we should deconstruct props in to meta, children, hint and input, so that it's clear what props are used

PR #40, opened 2018-01-30
Feature/update radio props

marksy:
- removed spacing from labels
- added spacing & overflow (for children that are floated) to gridRows
- added props to radio to support FinalForm


src/components/InputField/index.js
```diff
@@ -15,7 +15,7 @@ const InputField = ({ meta, children, hint, input }) => (
     <LabelText>{children}</LabelText>
     {hint && <HintText>{hint}</HintText>}
     {meta.touched && meta.error && <ErrorText>{meta.error}</ErrorText>}
-    <Input type="text" {...input} />
+    <Input {...input} />
```
penx: why get rid of the type?

marksy: i thought its not needed since we are setting a defaultProp of type='text'?

penx: oh are we, ok never mind

PR #43, opened 2018-01-31
option to disable flexbox for a column

marksy:
Suggestion to have columns `display: flex` (column) by default and have the ability to disable flexbox for a certain columns.

Currently:
1. columns are set to display:block
2. some children have floated elements (such as radios with prop of `inline`) which requires fixing the clearing (such as `overflow:hidden`, which isn't great if `drop-shadow` or `outline` is used outside of the box)
3. labels are being used in a block-level way for each line, which by default stretch to 100%
4. whitespace of labels are clickable, which can allow a checkbox to be accidentally clicked 
before

What I'm proposing to happen is:
1. have just the checkbox/label clickable (so the label is effectively inline, but clears each line as if it were block) (using `display:flex` and `align-items: flex-start`
after
2. floated elements dont float inside a flexbox, so to make them float we would pass a prop to set the display to block, allowing children to float
screen shot 2018-01-31 at 12 43 19

### DIFF
--- a/src/components/Label/index.js
+++ b/src/components/Label/index.js
@@ -10,6 +10,7 @@ const Label = glamorous.label(
     display: "flex",
     flexDirection: "column",
     boxSizing: "border-box",
+    width: "100%",
     [MEDIA_QUERIES.LARGESCREEN]: {
       maxWidth: SITE_WIDTH
     }
--- a/src/components/DateInput/__snapshots__/test.js.snap
+++ b/src/components/DateInput/__snapshots__/test.js.snap
@@ -28,7 +28,7 @@ exports[`DateInput matches snapshot: enzyme.mount 1`] = `
         >
           <glamorous(label)>
             <label
-              className="css-1dyhx6n"
+              className="css-16y5nda"
             >
               <glamorous(span)>
                 <span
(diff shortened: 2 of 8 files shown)


src/components/Label/index.js
```diff
@@ -10,6 +10,7 @@ const Label = glamorous.label(
     display: "flex",
     flexDirection: "column",
     boxSizing: "border-box",
+    width: "100%",
```
penx: can this change be made independent to the flex change? Or is it specifically tied to the css for the grid?

penx:
1. in terms of exported components, we don't want to go too deep in to providing a grid system for the parent application to use. If the parent application wants to use a React component based grid system, then there are projects that should support this https://github.com/roylee0704/react-flexbox-grid

2. internally in the application I understand we may want some patterns that are repeated for managing layout. I'm not convinced this should be a single grid component. I see more value in defining wrapper components where needed inside a component that are not exported.

3. I guess for instances where we are asking the parent application to provide a set of components as children this gets more complex. Perhaps you can provide example markup for what you're doing above? Are you expecting that a parent application would for example do:

```
<CheckboxGroup>
  <GridRow>
    <Checkbox />
  </GridRow>
  <GridRow>
    <Checkbox />
  </GridRow>
</CheckboxGroup>
```

?

penx:
Ok I think we're agreed on:

- where there are internal layout requirements, we will use internal layout/grid components and not export them
- parent projects can implement their own grid system or use an existing 3rd party solution

So think we can close this @marksy ?

PR #58, opened 2018-02-08
Feature/more tests for greater coverage

marksy:
- added more tests to components to gain 100% test coverage
- removed * imports from `govuk-colours` and added specific colour imports
- fixed glamorous props where they were failing 
- changed false ternary values to reference the object properties (rather than applying new values, see changes to _line 38_ `src/components/TextArea/index.js` for example)

### DIFF
--- a/src/components/Breadcrumb/index.js
+++ b/src/components/Breadcrumb/index.js
@@ -75,7 +75,7 @@ const Breadcrumb = ({ children }) => (
       {children.length && children.map ? (
         children.map((child, i) => (
           // eslint-disable-next-line react/no-array-index-key
-          <BreadcrumbListItem key={i}>{child}</BreadcrumbListItem>
+          <BreadcrumbListItem key={child.key || i}>{child}</BreadcrumbListItem>
         ))
       ) : (
         <BreadcrumbListItem>{children}</BreadcrumbListItem>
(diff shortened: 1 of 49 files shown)


src/components/Breadcrumb/index.js
```diff
@@ -57,23 +53,30 @@ const BreadcrumbListItem = glamorous.li({
...
     <BreadcrumbList>
+      {/*
+        the data of children doesnt have any IDs (or anything appropriate) so the last resort is to use the index for the key
+        https://reactjs.org/docs/lists-and-keys.html#keys
+        */}
       {children.length && children.map ? (
-        children.map(child => <BreadcrumbListItem>{child}</BreadcrumbListItem>)
+        children.map((child, i) => (
+          // eslint-disable-next-line react/no-array-index-key
+          <BreadcrumbListItem key={i}>{child}</BreadcrumbListItem>
```
penx: `key={child.key || i}` I think

src/components/TextArea/index.js
```diff
@@ -34,8 +34,8 @@ const TextareaField = glamorous.textarea(
       outlineOffset: 0
     }
   },
-  ({ errorText }) => ({
-    border: errorText ? `4px solid ${ERROR_COLOUR}` : `2px solid ${BLACK}`
+  ({ error }) => ({
+    border: error ? `4px solid ${ERROR_COLOUR}` : TextareaField.border
```
penx: I think

```
 `4px solid ${ERROR_COLOUR}` : undefined
```

would be best here, as it defaults it to whatever has been set above

PR #62, opened 2018-02-09
Bug/fix error variants

marksy:
- fixed components that had the error text not working
- updated snapshots
- fixed test
- changed false ternary values to `undefined`

### DIFF
--- a/src/components/Button/index.js
+++ b/src/components/Button/index.js
@@ -39,16 +39,16 @@ const Button = glamorous.button(
     }
   },
   ({ start }) => ({
-    fontWeight: start ? "700" : Button.fontWeight,
-    fontSize: start ? "24px" : Button.fontSize,
-    lineHeight: start ? "1.25" : Button.lineHeight,
-    padding: start ? "0.36842em 0.84211em 0.21053em" : Button.padding
+    fontWeight: start ? "700" : undefined,
+    fontSize: start ? "24px" : undefined,
+    lineHeight: start ? "1.25" : undefined,
+    padding: start ? "0.36842em 0.84211em 0.21053em" : undefined
   }),
   ({ iconUrl }) => ({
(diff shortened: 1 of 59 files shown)


src/components/Button/index.js
```diff
@@ -39,16 +39,16 @@ const Button = glamorous.button(
...
-    fontSize: start ? "24px" : Button.fontSize,
-    lineHeight: start ? "1.25" : Button.lineHeight,
-    padding: start ? "0.36842em 0.84211em 0.21053em" : Button.padding
+    fontWeight: start ? "700" : undefined,
+    fontSize: start ? "24px" : undefined,
+    lineHeight: start ? "1.25" : undefined,
+    padding: start ? "0.36842em 0.84211em 0.21053em" : undefined
   }),
   ({ iconUrl }) => ({
     backgroundImage: iconUrl ? `url( ${iconUrl})` : "",
```
penx: should this be `: "none",`?

penx: or undefined to not set it at all

src/components/Select/__snapshots__/test.js.snap
```diff
@@ -193,9 +193,6 @@ exports[`Select matches snapshot: enzyme.mount 1`] = `
   flex-direction: column;
   box-sizing: border-box;
   width: 100%;
-  border-left: ;
```
penx: I wonder if there's a way we can have a unit test that validates that css output validates

penx:
Can you add tests that make sure that, if this happens again, the tests will pick it up?

src/components/Select/__snapshots__/test.js.snap
marksy: would be good to get css linting working, but the plugins for `css-in-js`/glamorous aren't mature enough/working as expected

PR #67, opened 2018-02-13
Feature/beta banner

marksy:
- added phase badge (alpha/beta) component
- added anchor component
- added phase banner which includes badge, supporting text, link
- added tests

(alpha/beta) component should look like: https://govuk-elements.herokuapp.com/alpha-beta-banners
screen shot 2018-02-16 at 15 24 48

screen shot 2018-02-16 at 15 24 18


src/components/Anchor/index.js
```diff
@@ -0,0 +1,41 @@
+// https://govuk-elements.herokuapp.com/typography/
+
+import React from "react";
+import PropTypes from "prop-types";
+import glamorous from "glamorous";
+
+import { BLUE, PURPLE, YELLOW } from "govuk-colours";
+
+const AnchorInner = glamorous.a({
```
penx: Many applications will want to use a React Router `<Link />` component for anchors, but have it styled by glamorous.

We don't want to introduce a reliance on React Router, but this component should be configurable to support rendering/extending Link.

Could you look if there are similar styled-components/glamorous examples that do this? Or read the react router docs for Link and see if you can find any clear ways to do this?

Even better would be good to have a 'usage with react router' example in storybook.

src/components/Anchor/stories.js
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+
+import Anchor from ".";
+
+storiesOf("Anchor", module).add("Anchor default", () => (
+  <Anchor hyperLink="https://example.com">Anchor example</Anchor>
```
penx: it may be a case of just having an optional prop `component={Link}` but have a look if there's any docs/examples that would help

penx:
approved other than example of react router Link integration

penx:
When adding a new component, please provide a screenshot of the component in the PR description

PR #71, opened 2018-02-16
added anchor components for review

marksy:
new HOC which takes a component a returns another component with the props of the original component. This is to make style and make use of either anchor links and react-router links.

Link should look like: https://govuk-elements.herokuapp.com/typography#typography-links
screen shot 2018-02-16 at 15 21 13


src/components/asAnchor/test.js
```diff
@@ -0,0 +1,19 @@
...
+
+const wrapper = <AnchorTag href="https://example.com">Example</AnchorTag>;
+
+describe(asAnchor, () => {
+  it("renders without crashing", () => {
+    const div = document.createElement("div");
+    ReactDOM.render(wrapper, div);
+  });
+
+  it("matches wrapper snapshot", () => {
```
penx: please add at least one unit test that tests for expected functionality

src/components/asAnchor/test.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
```
penx: this file is an HOC not a component, please place in `src/hoc`

src/components/asAnchor/index.js
```diff
@@ -0,0 +1,46 @@
+import React from "react";
```
penx: this file is an HOC not a component, please place in `src/hoc`

.eslintrc
```diff
@@ -15,7 +15,7 @@
   },
   "overrides": [
     {
-      "files": [ "stories.js", "test.js", "src/setupTests.js", "scripts/createComponent.js", "src/stories/**" ],
+      "files": [ "stories.js", "test.js", "src/setupTests.js", "scripts/createComponent.js", "src/stories/**", "src/hoc/*/stories.js", "**/stories.js" ],
```
penx: I don't understand why `"src/hoc/*/stories.js"` is needed if we already have "**/stories.js"`

penx: similarly, `stories.js` should pick up all files with this name, doesn't it?

src/index.js
```diff
@@ -20,4 +21,9 @@ export { default as TextArea } from "./components/TextArea";
 export { default as MultiChoice } from "./components/MultiChoice";
 export { default as PhaseBadge } from "./components/PhaseBadge";
 export { default as PhaseBanner } from "./components/PhaseBanner";
+
+// Higher Level Components
```
penx: higher `order` not `level`

PR #74, opened 2018-02-20
Feature/babel 7

penx:
Use Babel 7 to do es and lib builds, rather than nwb.

This is because:

- nwb doesn't support a .babelrc file being present
- I want to write and run build scripts using babel-node, but this requires a .babelrc to be present
- Babel 7 now supports .babelrc.js and --config which makes it easy to have multiple or scripted configs


.babelrc
```diff
@@ -0,0 +1,20 @@
+{
```
penx: This is a copy of what nwb was doing for the lib build behind the scenes

config/babel/es.json
```diff
@@ -0,0 +1,19 @@
+{
```
penx: This is a copy of what nwb was doing for the es build behind the scenes

package.json
```diff
@@ -5,16 +5,17 @@
...
     "clean": "rimraf lib && rimraf es",
     "flow": "flow",
     "bundlesize": "bundlesize",
     "test": "CI=true npm run test:unit && npm run eslint && npm run flow",
-    "test:unit": "node scripts/test.js --env=jsdom",
-    "test:debug": "node  --inspect-brk scripts/test.js --runInBand --env=jsdom",
+    "test:unit": "jest --env=jsdom",
+    "test:debug": "node  --inspect-brk ./node_modules/.bin/jest --env=jsdom",
     "test:snapshot": "npm run test:unit -- -u",
-    "coverage": "npm run test:unit -- --coverage",
```
stevesims: did you intend to remove this?

without it, is it still possible to run a local test with a coverage report?

penx: yes, because our jest config in package.json:

```
    "coverageDirectory": "./coverage/",
    "collectCoverage": true,
```

means that test coverage is always obtained by default

package.json
```diff
@@ -5,16 +5,17 @@
     "start": "start-storybook -p 9009",
     "eslint": "eslint ./src",
     "eslint:fix": "eslint --fix ./src",
-    "build": "nwb build-react-component",
-    "build:dist": "nwb build-react-app",
+    "build": "npm run build:lib && npm run build:es",
+    "build:lib": "rimraf lib && babel src -d lib",
+    "build:es": "rimraf es && babel --config-file ./config/babel/es.json src -d es",
+    "build:dist": "webpack -p es/index.js dist/index.js",
```
stevesims: i'm wondering if this command should maybe be: `npm run build && webpack -p ...`

penx: yeah that was bothering me a bit too, it would make some builds take a little longer but I agree better to be safe - `npm run build:es && ...` should suffice

PR #79, opened 2018-02-26
Feature/api docs

penx:
Generate `API.md` by calling `npm run docs`.

This uses react-docgen.

Due to the default resolver of react-docgen not picking up Glamorous components, I've had to wrap some Glamorous only components with a JSX wrapper. I'd rather not have to do this but not sure of any alternative solution.

https://github.com/reactjs/react-docgen/issues/256

I'm also not sure the format of the current API.md is useful but opening PR for consideration.


scripts/docs/index.js
```diff
@@ -0,0 +1,70 @@
...
+async function generateApiForFiles(files) {
+  let md = "";
+  for (let i = 0; i < files.length; i += 1) {
+    const file = files[i];
+    if (shouldDocumentComponent(file)) {
+      md += await generateApiForFile(file);
+    }
+  }
+  return md;
+}
```
marksy: could possibly use `for of`?
``` 
for (const file of files) { 
  if (shouldDocumentComponent(file)) {
    md += await generateApiForFile(file);
  }
}
```

scripts/docs/loadExample.js
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  if (typeof Component === "object") {
+    Component = Component.default;
+  }
+
+  if (!Component) {
+    console.log(chalk.red("Error loading component:"), file);
+  }
+
```
marksy: is it necessary to have two `if (!Component)`?

scripts/docs/markdown/generateProps.js
```diff
@@ -0,0 +1,19 @@
+import stringOfLength from "../stringOfLength";
+import generateProp from "./generateProp";
+
+export default function generateProps(props) {
+  const title = "Props";
+
+  if (!props) {
+    return "";
+  }
+
```
marksy: maybe picky, but perhaps move the `if !props, return` to the top to avoid the `const title = "Props";` from being defined

marksy: _unless need the title there anyway_

scripts/docs/index.js
penx: I did originally but got this from eslint:

> iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations. (no-restricted-syntax)

discussion https://github.com/airbnb/javascript/issues/1271

scripts/docs/stringOfLength.js
```diff
@@ -0,0 +1,7 @@
+export default function stringOfLength(string, length) {
+  let newString = "";
+  for (let i = 0; i < length; i += 1) {
+    newString += string;
+  }
```
marksy: could maybe use `for of` here too?

scripts/docs/loadExample.js
penx: second is "if there still isn't a component after loading the default", so yes

src/components/Button/index.js
```diff
@@ -6,14 +6,16 @@
...
+
 import {
   BUTTON_COLOUR,
   BUTTON_COLOUR_DARKEN_15,
   WHITE,
   YELLOW
 } from "govuk-colours";
 
-const Button = glamorous.button(
+const GButton = glamorous.button(
```
marksy: was this supposed to be renamed?

scripts/docs/loadExample.js
penx: 1. try load example.js
2. if this didn't load, index.js
3. if this isn't a component, look for default export
4. if still no component then log

src/components/Button/index.js
marksy: ah looks like it, although the naming is confusing. Maybe `GlamorousButton`? 👨‍💻

penx: yeah, react-docgen doesn't pick up glamorous components so I'm having to wrap them with a JSX component. I renamed the inner components to start with a G for Glamorous

scripts/docs/stringOfLength.js
penx: same as https://github.com/penx/govuk-react/pull/79#discussion_r170896895

scripts/docs/index.js
marksy: ok fair play

scripts/docs/loadExample.js
marksy: ah true cool

scripts/docs/stringOfLength.js
marksy: ok

src/components/Button/index.js
marksy: okay might be good to document that either inline comments or in the readme

penx: it's in the PR description and I've created an issue here https://github.com/penx/govuk-react/issues/81

penx: I agree it's not great, I just didn't want to spend too much time on documentation at this point, will hopefully revisit soon, maybe there will be progress on glamorous + react-docgen compatibility in the meantime and we can just undo this.

marksy: ok sounds good

PR #82, opened 2018-02-27
Pagination component

marksy:
## A pagination (previous/next) component with HOC paginationItem (either previous or next)
Can use either Link or HTML anchor tag.
Default:
screen shot 2018-02-26 at 13 18 01

With page numbers:
screen shot 2018-02-26 at 13 17 50

With labels:
screen shot 2018-02-26 at 13 18 09

Also fixed a minor bug with create-react script (showed a successful log message regardless of success)


src/icons/ArrowRight/index.js
```diff
@@ -0,0 +1,29 @@
...
+    xmlns="http://www.w3.org/2000/svg"
+    x="0px"
+    y="0px"
+    width="17"
+    height="13"
+    style={{ display: "block" }}
+    fill={colour}
+    viewBox="0 0 17 13"
+  >
+    <title>Next</title>
```
penx: text in svgs should be props with a defaultprop

src/icons/ArrowRight/index.js
```diff
@@ -0,0 +1,29 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+const ArrowRight = ({ colour }) => (
```
penx: may be good practice to pass rest props to the svg node on all svgs `({ colour, ...rest })`

``` <svg
     version="1.1"
     xmlns="http://www.w3.org/2000/svg"
     x="0px"
     ...
     viewBox="0 0 17 13"
     {...rest}
   >```

File: src/icons/ArrowRight/index.js [hunk 3]
```diff
@@ -0,0 +1,29 @@
...
+  <svg
+    version="1.1"
+    xmlns="http://www.w3.org/2000/svg"
+    x="0px"
+    y="0px"
+    width="17"
```
penx: do we have to set a width and height or can we make it fill 100% of the parent element? And assume the parent element is set to be 17x13?

src/hoc/asPaginationItem/index.js
```diff
@@ -0,0 +1,136 @@
...
+  textDecoration: "underline",
+  [MEDIA_QUERIES.LARGESCREEN]: {
+    fontSize: FONT_SIZE.SIZE_16,
+    lineHeight: LINE_HEIGHT.SIZE_16
+  }
+});
+
+const asPaginationItem = AnchorType => {
+  const component = props => (
+    <PaginationWrapper
```
penx: Do we have to use a span wrapper? e.g. with `glamorous(AnchorType)`

I think we discussed this for the anchor hoc but not sure if we resolved it.

Maybe we should add a new issue to look in to this?

src/icons/ArrowRight/index.js
marksy: yeah i'll make it 100% x/y

src/hoc/asPaginationItem/index.js
marksy: okay. I spent a bit of time last week trying to style the hoc but didnt have much luck. i felt that wrapping a glamorous tag around was a quick and didnt seem to have many issues (apart from an extra tag)

src/hoc/asPaginationItem/index.js
```diff
@@ -0,0 +1,140 @@
...
+  lineHeight: LINE_HEIGHT.SIZE_14,
+  textDecoration: "underline",
+  [MEDIA_QUERIES.LARGESCREEN]: {
+    fontSize: FONT_SIZE.SIZE_16,
+    lineHeight: LINE_HEIGHT.SIZE_16
+  }
+});
+
+const asPaginationItem = AnchorType => {
+  const component = props => (
```
penx: component is a component so should be CamelCase,

e.g.

```
const asPaginationItem = Composed => {
   const PaginationItem = props => (
```

PR #88, opened 2018-02-28
Feature/api docs table

marksy:
Reformatted the API docs to show a table of data rather than paragraphs

Before:
screen shot 2018-02-28 at 19 34 17

After:
screen shot 2018-02-28 at 19 33 12


penx:
Awesome

PR #89, opened 2018-03-01
Feature/remove prettier

penx:
remove prettier


src/components/BackLink/index.js
```diff
@@ -1,48 +1,48 @@
...
-    borderStyle: "solid",
-    borderColor: "transparent",
-    clipPath: "polygon(0% 50%, 100% 100%, 100% 0%)",
-    borderWidth: "5px 6px 5px 0",
+    borderStyle: 'solid',
+    borderColor: 'transparent',
+    clipPath: 'polygon(0% 50%, 100% 100%, 100% 0%)',
+    borderWidth: '5px 6px 5px 0',
     borderRightColor: `${BLACK}`,
     content: "''",
```
marksy: situations like this should really be inversed like `'""'`

PR #102, opened 2018-03-05
unit tests for constants exports

penx:
unit tests for constants exports

### DIFF
--- a/packages/constants/src/test.js
+++ b/packages/constants/src/test.js
@@ -1,0 +1,19 @@
+// every export from constants should be a truthy (string, number, bool or plain non empty object)
+// no functions, no undefined
+import * as exports from '.';
+
+const okTypes = ['number', 'string', 'boolean', 'object'];
+
+describe('Constants', () => {
+  for (let n = 0; n < Object.keys(exports).length; n += 1) {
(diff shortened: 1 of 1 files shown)


packages/constants/src/test.js
```diff
@@ -0,0 +1,20 @@
...
+  for (let n = 0; n < Object.keys(exports).length; n += 1) {
+    const key = Object.keys(exports)[n];
+
+    it(`exports ${key} as valid type`, () => {
+      const value = exports[key];
+      // TODO: this would fail if we wanted to export a constant that was `false`
+      // not sure how to do "expect one of" in Jest so leaving for now
+      // as it's possible we'd never need a false constant export
+      expect(value).toBeTruthy();
+      expect(okTypes.includes(typeof value)).toBe(true);
```
stevesims: i'd have thought this line alone would be sufficient...

if you really want a "one of" tho, i'd expect the following to work:
```js
expect(value || okTypes.includes(typeof value)).toBeTruthy();
```

penx: addressed in https://github.com/penx/govuk-react/pull/102/commits/2bb0b47abf4c105740f5042da87de84a766a2a29

PR #104, opened 2018-03-05
Feature/codecoverage missing components

penx:
- snapshot tests for input, input-text, label, multichoice
- hoc and icons export type test


penx:
if you can find an eslint config setting that would do it, I'm happy to add it

PR #117, opened 2018-03-07
added example of last breadcrumb being not a link

### LINKED ISSUE #97: Breadcrumb where last item is *not* a link?
Usual usage of breadcrumbs has the last item in the list being the current page name/title, and thus not a link.

The current storybook examples for the `Breadcrumb` component don't follow this pattern, but instead have every item as a link.

If it's possible to do conventional breadcrumbs (with last item being just the page title) with the `Breadcrumb` component we should update the storybook examples accordingly.

If it's not possible we should adjust the component and then also adjust the storybook examples.  :grin:

marksy:
Fixes #97 

- added examples to show the current page (not being a link)


marksy:
there were no modifications to the component src, just the story example

PR #118, opened 2018-03-07
added ordered list, fixed unordered list padding (fixes #110)

### LINKED ISSUE #110: List styles don't match GDS
The style of the `UnorderedList` component doesn't match GDS.

The GDS style does not proscribe a width, plus it has padding and margin values which are have not been included.

Additionally we don't currently have an `OrderedList` component, or support for numbered lists.

marksy:
Fixes #110 

New component `ordered-list`
screen shot 2018-03-07 at 14 37 36

---

Fixed `unordered-list` padding, #110
before:
screen shot 2018-03-07 at 14 39 25

after:
screen shot 2018-03-07 at 14 39 43

### DIFF
--- a/components/list-item/src/index.js
+++ b/components/list-item/src/index.js
@@ -6,15 +6,16 @@ import {
   LINE_HEIGHT,
   MEDIA_QUERIES,
   NTA_LIGHT,
+  SPACING,
 } from '@govuk-react/constants';
 
 const GListItem = glamorous.li({
-  boxSizing: 'border-box',
   fontFamily: NTA_LIGHT,
   fontWeight: 400,
   textTransform: 'none',
   fontSize: FONT_SIZE.SIZE_14,
   lineHeight: LINE_HEIGHT.SIZE_14,
+  marginBottom: SPACING.SCALE_1,
   width: '100%',
   [MEDIA_QUERIES.LARGESCREEN]: {
     fontSize: FONT_SIZE.SIZE_16,
--- a/components/list-item/src/index.js
+++ b/components/list-item/src/index.js
@@ -15,11 +15,11 @@ const GListItem = glamorous.li({
   textTransform: 'none',
   fontSize: FONT_SIZE.SIZE_14,
   lineHeight: LINE_HEIGHT.SIZE_14,
-  marginBottom: SPACING.SCALE_1,
   width: '100%',
   [MEDIA_QUERIES.LARGESCREEN]: {
     fontSize: FONT_SIZE.SIZE_16,
     lineHeight: LINE_HEIGHT.SIZE_16,
+    marginBottom: SPACING.SCALE_1,
   },
 });
 
--- a/components/ordered-list/src/index.js
+++ b/components/ordered-list/src/index.js
@@ -6,7 +6,6 @@ import glamorous from 'glamorous';
 
 const GOrderedListInner = glamorous.ol(
   {
-    boxSizing: 'border-box',
     margin: 0,
     padding: 0,
     paddingLeft: '20px',
--- a/components/ordered-list/src/index.js
+++ b/components/ordered-list/src/index.js
@@ -4,7 +4,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import glamorous from 'glamorous';
 
-const GUnorderedList = glamorous.ul(
+const GOrderedListInner = glamorous.ol(
   {
     boxSizing: 'border-box',
     margin: 0,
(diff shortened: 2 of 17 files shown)


scripts/createComponent.js
```diff
@@ -3,22 +3,22 @@
...
-/* eslint no-console: 0 */
-const createFolder = (fName) => {
-  mkdirp(fName, (err) => {
-    if (err) {
-      console.error(err);
-    }
-  });
-  mkdirp(`${fName}/src`);
-};
+// /* eslint no-console: 0 */
```
penx: why comment this block out?

penx: I think it can be safely deleted

components/ordered-list/src/stories.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import ListItem from '@govuk-react/list-item';
+
+import OrderedList from '.';
+
+storiesOf('OrderedList', module).add('OrderedList default', () => (
+  <OrderedList>
+    <ListItem>OrderedList example</ListItem>
```
stevesims: it might be an idea to add a second `ListItem` to these examples so that one gets to better see how lists are rendered, including spacing between elements

components/ordered-list/src/index.js
```diff
@@ -0,0 +1,35 @@
+// https://govuk-static.herokuapp.com/component-guide/government_navigation
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import glamorous from 'glamorous';
+
+const GOrderedListInner = glamorous.ol(
+  {
+    boxSizing: 'border-box',
```
stevesims: i think it may be better to put in font-related style definitions here

this is how govuk-frontend does things, and the other older govuk styling too.

arguably the `ListItem` component shouldn't need much styling - govuk-frontend just has margin-bottom applied to `li` items (with media queries)

components/unordered-list/src/stories.js
```diff
@@ -1,25 +1,26 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
+import ListItem from '@govuk-react/list-item';
 
 import UnorderedList from '.';
 
 storiesOf('UnorderedList', module).add('UnorderedList default', () => (
   <UnorderedList>
-    <li>UnorderedList example</li>
+    <ListItem>UnorderedList example</ListItem>
```
stevesims: As per comments against the `ordered-list` stories, would be nice to see at least 2 `ListItem` lines here

components/ordered-list/src/index.js
stevesims: this comment of course also applies to the `unordered-list` component too 😀

components/list-item/src/index.js
```diff
@@ -6,15 +6,16 @@ import {
...
 } from '@govuk-react/constants';
 
 const GListItem = glamorous.li({
-  boxSizing: 'border-box',
   fontFamily: NTA_LIGHT,
   fontWeight: 400,
   textTransform: 'none',
   fontSize: FONT_SIZE.SIZE_14,
   lineHeight: LINE_HEIGHT.SIZE_14,
+  marginBottom: SPACING.SCALE_1,
```
stevesims: i think there's actually a media query associated with this in govuk-frontend - but it's not clear to me exactly how to read how that stuff works

stevesims: yeah - the actual CSS that's kicking in is: ```css
@media (min-width: 40.0625em)
.govuk-list > li, .govuk-prose-scope ul > li, .govuk-prose-scope ol > li {
    margin-bottom: 5px;
}
```

File: components/list-item/src/index.js [hunk 6]
```diff
@@ -6,15 +6,16 @@ import {
   LINE_HEIGHT,
   MEDIA_QUERIES,
   NTA_LIGHT,
+  SPACING,
 } from '@govuk-react/constants';
 
 const GListItem = glamorous.li({
-  boxSizing: 'border-box',
   fontFamily: NTA_LIGHT,
```
stevesims: still think i'd rather see the font stuff removed from here and placed into the `OrderedList` and `UnorderedList` components - but not going to block this merge for that

penx: I think it should be in both places but not really needed right now

penx: had a discussion with @stevesims in the office, I think in summary this is - we are using these components in an environment (DOM) where CSS cascades down. However if we were to have react native compatibility in the future I'm not sure these styles would cascade down. Part of the thinking around CSS in JS is complete encapsulation of the styles for a component so we shouldn't rely on a parent component's styles cascading down.

marksy: okay i'll take out of `ul`/`ol` - `ListItem` which is the next child from those, has font styling etc

marksy: okay might just leave as-is for now, if thats okay and if and when we start using react native we can address then. 

I'm not sure having font styling in both the `ul/ol` and the `ListItem` is useful, as the li the only child `ul/ol` can have iirc

PR #126, opened 2018-03-12
Feature/modular svg

### LINKED ISSUE #84: Make svg icons more modular
Currently each svg has the same tag & attributes so would be beneficial to make a resuable component which passes in props. Just the path would be different for each icon

marksy:
- created an SVG component wrapper that takes props (title, colour, ...rest) and children (svg path)
- updated existing svg icons
fixes #84


packages/icons/src/SVGBase/index.js
```diff
@@ -0,0 +1,35 @@
...
+}) => (
+  <svg
+    version="1.1"
+    xmlns="http://www.w3.org/2000/svg"
+    x="0px"
+    y="0px"
+    width="100%"
+    height="100%"
+    style={{ display: 'block' }}
+    fill={colour}
```
penx: where we are passing properties down as-is, unless there is a reason not to, I would persist the prop name, so in this case I would call the prop `fill`

PR #127, opened 2018-03-12
Fix for duff breadcrumb children

### LINKED ISSUE #123: Breadcumb enhancement: strip empty/duff children
When attempting to use the `Breadcrumb` component I had code like the following:
```jsx
<Breadcrumb>
  <Link to="/">Home</Link>
  {crumbs}
  {currentPage}
</Breadcrumb>
```

`crumbs` here was the result of mapping a different array of pages to create links back to them.  This works fine when there are actually pages given, but when the list is empty you get an empty crumb, and see something like: `Home > > currentPage`

I worked around this by building up an array of children instead, and using `<Breadcrumb children={crumbs} />`

It may be a good enhancement to have the `Breadcrumb` strip out "empty" children.

marksy:
fixes #123 

When the breadcrumb was being passed in as null/empty array it was showing a "ghost" breadcrumb
*before:*
screen shot 2018-03-12 at 11 01 04

*after:*
screen shot 2018-03-12 at 10 59 02

### DIFF
--- a/components/breadcrumb/src/index.js
+++ b/components/breadcrumb/src/index.js
@@ -70,7 +70,8 @@ const Breadcrumb = ({ children }) => (
     <BreadcrumbList>
       {children.length && children.map ? (
         children.map((child, i) => (
-          <BreadcrumbListItem key={child.key || i}>{child}</BreadcrumbListItem>
+          child.length || child.props ? <BreadcrumbListItem key={child.key || i}>{child}</BreadcrumbListItem>
+ : null
         ))
       ) : (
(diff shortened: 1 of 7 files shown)


penx:
Bugfixe PRs should include a unit test

penx:
I think where we have nested components (breadcrumbs, tables, lists) we should be using the React context API

This is going to be in React 16.3

https://medium.com/dailyjs/reacts-️-new-context-api-70c9fe01596b

https://github.com/facebook/react/issues/12152

Should probably just mark this as a TODO for now

PR #129, opened 2018-03-13
Feature/ Related Items

marksy:
- New `Related Items` component
screen shot 2018-03-13 at 13 59 50

- updated Unordered List component to support recomposing (restyling)


components/unordered-list/src/index.js
```diff
@@ -37,17 +37,19 @@ const GUnorderedList = glamorous.ul(
 
 // listStyleType: normal HTML property values are used here
 // e.g., listStyleType="circle", listStyleType="upper-alpha", listStyleType="none"
-const UnorderedList = ({ children, listStyleType }) => (
-  <GUnorderedList listStyleType={listStyleType}>{children}</GUnorderedList>
+const UnorderedList = ({ children, listStyleType, className }) => (
+  <GUnorderedList listStyleType={listStyleType} css={[className]}>{children}</GUnorderedList>
```
penx: as discussed this morning, not sure we should be doing this by default, only when there's a specific case for allowing any css override

PR #131, opened 2018-03-14
Allow recomposition/restyling of existing components

marksy:
- Added css prop to components that allows pass-through of styles to be able to override styles of a component, should it need it
- Added an HoC to give an example of how to override the component, not intended for export
- Added a test for the HoC to make sure it does what it's supposed to
- Fixed a couple of issues
  + `null` where should have been `undefined`
  + textarea was consuming all props, instead of just input props

Usage of the change would be as follows:
```
import BackLink from '@govuk-react/back-link';
const NewBackLink  = glamorous(BackLink)({
  backgroundColor: "blue",
  color: "white",
});
```

See [glamorous website](https://glamorous.rocks/basics/) "Overriding component styles" section (however it doesn't clearly explain the method, so i [wrote one](https://medium.com/@marksy/re-styling-a-compose-glamorous-component-b6c6aa728a3b) based on [KCD's tweet to me](https://twitter.com/kentcdodds/status/973531839708127232))

### DIFF
--- a/components/back-link/src/__snapshots__/test.js.snap
+++ b/components/back-link/src/__snapshots__/test.js.snap
@@ -49,6 +49,11 @@ exports[`Back Link matches wrapper snapshot: wrapper mount 1`] = `
   goBack={[Function]}
 >
   <glamorous(button)
+    css={
+      Array [
+        undefined,
+      ]
+    }
     onClick={[Function]}
   >
     <button
--- a/components/back-link/src/index.js
+++ b/components/back-link/src/index.js
@@ -45,8 +45,8 @@ const Anchor = glamorous.button({
   },
 });
 
-const BackLink = ({ children, goBack }) => (
-  <Anchor onClick={goBack}>{children}</Anchor>
+const BackLink = ({ children, goBack, className }) => (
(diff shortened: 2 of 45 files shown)


penx:
We need to be careful about this:

- what styles would be ok to live on the same element?
- are there occasions having additional styles on a wrapping element would suffice?
- if we want to support particular overrides such as background colour or text colour, could these be props that are passed in?

There are a lot of things that could get messy by allowing anything to be overridden.

The nice thing about BEM > CSS Modules > CSS in JS is that it's predictable and styles are self contained.

This makes the styles that are applied to an element unpredictable.

e.g. this element looks incorrect, which CSS is to blame? What needs changing?

e.g. a component gets updated and breaks components that are using it by applying CSS to the same element

e.g. a component uses color for text, which is overridden elsewhere, but then a code update puts the text in a `:before{content:` and the overriding CSS no longer works. But the ability to change the colour was never a documented feature of the component... so is this ok?

marksy:
yeah agreed, but my reasoning was to have the ability recompose a component if needs be, rather than disallow.. but yeah up for a discussion

PR #145, opened 2018-03-19
missing component examples

marksy:
- adding missing component examples
- updated searchbox to show correctly 
- removed Grid & Layout images, and added them to git ignore


scripts/createComponent.js
```diff
@@ -34,6 +34,7 @@ const packageJson = () => {
   "name": "@govuk-react/${componentFolderName}",
   "version": "${version}",
   "dependencies": {
+    "@govuk-react/constants": "${version}",
```
penx: I think `^${version}` may be better, as it's ok to use an updated version of constants with non breaking changes

penx: but similarly, there may not be a version of constants matching the current lerna version, which I think is why I removed this, it's a bit tricky to know in advance what version would be needed

API.md
```diff
@@ -188,10 +192,10 @@ ListNavigation
 ![Component Image](./docs/ListNavigation.png)
 
 
-Prop | Required | Default | Type | Description
```
penx: Is this white space change because you opened the file in your editor and then saved it?

Perhaps fix the white space changes in the generator script too, so that next time someone generates this file it doesn't put all the white space back in again?

scripts/createComponent.js
marksy: should i take out this line?

penx: I reckon so

API.md
marksy: yeah

PR #147, opened 2018-03-19
remove version numbers from root package.json

penx:
Remove version numbers from root package.json, so that outdated version references are not left behind after a `lerna publish`, see:

https://github.com/lerna/lerna/issues/1335


marksy:
can you please resolve the conflict `yarn.lock`

PR #148, opened 2018-03-20
Feature/custom error colour input

penx:
Allow input to have a custom error colour

### DIFF
--- a/components/input/src/index.js
+++ b/components/input/src/index.js
@@ -33,7 +33,7 @@ const GInput = glamorous.input(
     },
   },
   ({ error, errorColour }) => ({
-    border: error ? `4px solid ${errorColour}` : undefined,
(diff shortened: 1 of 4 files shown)


components/input/src/index.js
```diff
@@ -32,8 +32,8 @@ const GInput = glamorous.input(
       outlineOffset: 0,
     },
   },
-  ({ error }) => ({
-    border: error ? `4px solid ${ERROR_COLOUR}` : undefined,
+  ({ error, errorColour }) => ({
```
stevesims: `Colour` here kinda grates with me a bit - as this is styling i'd personally prefer `Color`
maybe we should support both, and `||` them?

penx: I agree, I was being consistent with what we've used previously, but I'd be happy to standardize on color

PR #151, opened 2018-03-20
Feature/component readme generation

penx:
- docs script for all components
- don't save doc image to disk if it's 0 width or height
- don't include doc image in markdown if it doesn't exist on disk
- npm run docs:components, this duplicates images at the root for now, may dedupe in a future PR
- don't include images in API.md if they don't exist on disk, (re run `npm run docs`)


stevesims:
This MR shows we've got a few components that aren't fully documented e.g. `ErrorText` and `HintText`, which should say they accept `children` - but that's a separate issue (see #153)

PR #162, opened 2018-03-22
Bugfix/eslint

penx:
Eslint errors weren't being picked up since move to monorepo.


package.json
```diff
@@ -5,8 +5,8 @@
     "bootstrap": "lerna bootstrap",
     "docs": "./packages/api-docs/bin/doc-component.js './components/**/lib/index.js' './API.md'",
     "docs:components": "lerna run docs --parallel",
-    "eslint": "eslint ./src",
-    "eslint:fix": "eslint --fix ./src",
+    "eslint": "eslint ./packages/**/src/**.js",
```
stevesims: just wondering why this isn't including `components` as well as `packages` like the following line does?

packages/icons/src/Spinner/index.js
```diff
@@ -22,6 +22,7 @@ const Rect = glamorous.rect(animationStyles);
 const Spinner = ({ fill, title, ...rest }) => (
   <SVG viewBox="-50 -50 100 100" preserveAspectRatio="xMidYMid meet" title={title} fill={fill} {...rest}>
     {Array(12).fill(1).map((el, i) =>
+      /* eslint-disable react/no-array-index-key */
```
stevesims: this should be `eslint-disable-next-line`

package.json
penx: that's a mistake!

PR #179, opened 2018-03-28
Loading box component

marksy:
- new loading-box
- unit tests

**Fixed sized loaders**
Create a component called Spinner that could be used as follows:
```
<Placeholder
fallback={<Spinner size="large" />}
>
<SomeContent />
</Placeholder>
```

This is likely the existing just the loading icon but at 3 preset sizes: small, medium, large. The sizes of small, medium and large loaders need to be decided by the team. A small loader should be appropriately sized to fit in a button. A large loader could be used to show an entire page is loading. A medium loader would be used e.g. to show a 500px x 500px area of a page is loading.

**LoadingBox**
Create a LoadingBox component that can be used like this
```
<LoadingBox loading={true} color="rgba(4,3,2,1)" timeIn={1000} timeOut={1000}>
  <AnyReactNodes />
  <AnyReactNodes />
</LoadingBox >
```

When the component mounts and loading is true:
* children are rendered
* all children should be covered by a 50% grey at 50% opacity (colour is configurable prop)
* anything positioned outside this area is hidden (overflow hidden)
* a large loading spinner is positioned with its centre point in the vertical and horizontal centre of the container (component to use is a configurable prop)
* if the height of the container is greater than half height of the viewport then the centre of the loader icon is positioned at a quarter of the viewport height from the top of the container (max-viewport-percentage is a configurable prop defaults to 0.25)

When the component mounts and loading is false (default)
* the ValidatedContent does nothing, just shows children

When component has previously mounted and loading switches from true to false
the loading icon and grey area fade out over 0.3 seconds (time-out is configurable prop)
`overflow: hidden` is disabled after animation is completed

When component has previously mounted and loading switches from false to true
* the loading icon and grey area fade in over 0.3 seconds (time-in is configurable prop)
* overflow hidden is set as soon as animation starts


components/loading-box/src/index.js
```diff
@@ -0,0 +1,166 @@
...
+};
+
+LoadingBox.propTypes = {
+  children: PropTypes.node.isRequired,
+  spinnerColor: PropTypes.string,
+  // hex colour 3 or 6 characters with or without hash
+  backgroundColor: PropTypes.string,
+  backgroundColorOpacity: PropTypes.number,
+  loading: PropTypes.bool,
+  // number in milliseconds
```
penx: this text goes in to the docs, may be worth repeating for both timeIn and timeOut, also extend to say:

// length of fade in animation in milliseconds
// length of fade out animation in milliseconds

components/loading-box/src/index.js
```diff
@@ -0,0 +1,166 @@
...
+        <Overlay />
+      </Innerwrap>
+    </CSSTransition>
+    {children}
+  </Wrapper>
+);
+
+LoadingBox.defaultProps = {
+  spinnerColor: WHITE,
+  backgroundColor: '#808080',
```
penx: 50% gray ✅

marksy: * grey

penx: did we discuss making this white by default?

marksy: yeah but you said this was configurable in with knobs anyway

components/loading-box/src/stories.js
```diff
@@ -0,0 +1,68 @@
...
+  <LoadingBox
+    loading={boolean('loading', false)}
+    backgroundColor={text('backgroundColor', '#808080')}
+    timeIn={number('timeIn', 200)}
+    timeOut={number('timeOut', 200)}
+    backgroundColorOpacity={number('backgroundColorOpacity', 0.7)}
+    spinnerColor={text('spinnerColor', '#000')}
+  >
+  <h1 style={{ marginTop: 0 }}>Toggle settings under 'knobs'</h1>
+  <p>Ut enim ad minim veniam, quis laboris nisi ut aliquip ex ea commodo consequat.</p>
```
penx: Please can you include a few more elements as child components of the example? At least a button and an input field.

Please can you also make sure the content looks GDS-ish, i.e. no times new roman text

components/loading-box/src/index.js
marksy: i will change to white

package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "govuk-react",
   "scripts": {
+    "yo": "yarn && yarn bootstrap && yarn build && echo WATCH && yarn watch",
```
penx: yarn will yarn bootstrap
yarn watch will yarn build

so you should just need `yarn && echo WATCH && yarn watch`

package.json
```diff
@@ -104,6 +105,7 @@
     "flow-bin": "^0.68.0",
     "glamor": "^2.20.40",
     "glamorous": "^4.11.2",
+    "hex-rgb": "^2.0.0",
```
penx: this shouldn't be needed at the root

package.json
```diff
@@ -115,6 +117,7 @@
     "react-final-form": "^3.1.0",
     "react-router-dom": "^4.2.2",
     "react-test-renderer": "^16.2.0",
+    "react-transition-group": "^2.2.1",
```
penx: this shouldn't be needed at the root

components/loading-box/src/index.js
```diff
@@ -0,0 +1,170 @@
+// TODO INSERT A COMMENT REFERENCE TO EXTERNAL URL IF POSSIBLE
```
penx: add url to existing loading box component

components/loading-box/src/index.js
```diff
@@ -0,0 +1,170 @@
...
+}) => ({
+  position: 'absolute',
+  height: '100%',
+  top: 0,
+  right: 0,
+  left: 0,
+  bottom: 0,
+  display: 'flex',
+  justifyContent: 'center',
+  '> svg': {
```
penx: I'm having a think about this... thoughts to come

penx: This, and `> div` is assuming knowledge of the implementation of the child components. If the child components elements were to change, this would break.

Please read [Component as a selector](https://glamorous.rocks/examples/#component-as-a-selector) example in the Glamorous docs, specifically:

1. how you could use a component selector instead of `> svg` 
2. the point that " This is actually intentional because it's not a pattern you should generally follow (one of the nice parts of css-in-js is that it allows you to not worry about the cascade of CSS)"

Further to (2), is there any way we could achieve the same without styling an external component? E.g. creating a Glamorous component internal to this file and styling that with props?

penx: I would prefer not having a cascading style, but could live with a component as a selector 😀

components/loading-box/src/example.js
```diff
@@ -0,0 +1,4 @@
+import React from 'react';
+import LoadingBox from '.';
+
+export default () => <LoadingBox>LoadingBox example</LoadingBox>;
```
penx: The example is used to generate the screenshot and should probably have the `loading` prop

packages/icons/src/SVGBase/index.js
```diff
@@ -11,6 +11,7 @@ const SVG = ({
     y="0px"
     width="100%"
     height="100%"
+    className={className}
```
penx: className would already have been picked up and passed on by `...rest` so this change doesn't do anything, but perhaps good to have in for clarity anyway?

components/loading-box/src/index.js
```diff
@@ -0,0 +1,170 @@
...
+    {children}
+  </Wrapper>
+);
+
+LoadingBox.defaultProps = {
+  spinnerColor: BLACK,
+  backgroundColor: WHITE,
+  backgroundColorOpacity: 0.5,
+  loading: false,
+  timeIn: 200,
```
penx: I think the default timeIn should be longer than the default timeOut

components/loading-box/src/stories.js
```diff
@@ -0,0 +1,95 @@
...
+      <Button start>Continue</Button>
+    </div>
+  </LoadingBox>
+));
+
+stories.add('LoadingBox (long)', () => (
+  <LoadingBox
+    loading={boolean('loading', false)}
+    backgroundColor={text('backgroundColor', '#fff')}
+    timeIn={number('timeIn', 300)}
```
penx: if this is:

```
    timeIn={number('timeIn', undefined)}
    timeOut={number('timeOut', undefined)}
```

Will it default to the defaultProps for these numbers? If so can you do that? If not, can you set the defaults to be the same as the default props? See previous comments on default timeIn being longer than default timeOut

penx:
When loading is true, and the overlay is visible, when do not expect the user to be able to interact with the children elements.

They can however still be tabbed to.

We should do some research into aria tags and any other techniques to mark this content as disabled.

e.g.

`user-select: none` https://caniuse.com/#feat=user-select-none

`aria-disabled="true"` https://developer.mozilla.org/en-US/docs/Learn/Accessibility/WAI-ARIA_basics

We don't need to do this now, just worth having a quick think about it.

marksy:
> They can however still be tabbed to.

we can also disable field elements to show `readOnly` sort of state

PR #184, opened 2018-04-03
add child proptypes

### LINKED ISSUE #153: Ensure docs show `children` prop when applicable
Some components like `ErrorText`, `HintText` and `LabelText` accept `children`, but this isn't in their documentation.  This is presumably a missing `propTypes` definition, which we should correct

marksy:
fixes #153 

- added propTypes to ErrorText, LabelText, HintText
- fixed LoadingBox doc image


components/error-text/src/index.js
```diff
@@ -23,5 +24,10 @@ const GErrorText = glamorous.span({
   },
 });
 
-const ErrorText = props => <GErrorText {...props} />;
+const ErrorText = ({ children }) => (<GErrorText>{children}</GErrorText>);
```
stevesims: this is no longer spreading props, so we'd lose the ability to set stuff like `id` or other html attributes on this element

we should either revert the change to this line (as the previous version would still work fine), or make this line:
`const ErrorText = ({ children, ...props }) => (<GErrorText {...props}>{children}</GErrorText>);`

components/hint-text/src/index.js
```diff
@@ -21,6 +22,13 @@ const GHintText = glamorous.span({
   },
 });
 
-const HintText = props => <GHintText {...props} />;
+const HintText = ({ children }) => (
```
stevesims: as per comment against `ErrorText`, this change should either be reverted or changed to spread `props`

(spreading `props` on `GHintText` as per the previous version of this line passes through `children` as desired, without the need to explicitly set the children)

components/label-text/src/index.js
```diff
@@ -27,6 +28,13 @@ const GLabelText = glamorous.span(
   }),
 );
 
-const LabelText = props => <GLabelText {...props} />;
+const LabelText = ({ children }) => (
```
stevesims: same comment as per `HintText` and `ErrorText` :grin:

components/error-text/src/index.js
marksy: Cool thanks

PR #185, opened 2018-04-03
Whitespace / marginBottom hoc

### LINKED ISSUE #173: White space
All components should have default margin-bottom such that when they are used together they space themselves out automatically.

`marginBottom` should be an optional prop on these components that uses a number 0 to 9 representing the spacing units listed at:

https://govuk-design-system-production.cloudapps.digital/styles/spacing/

(to discuss - `:last-child` components should always have a margin bottom of 0?)

We will likely use Cosmos' implementation as a guide for handling this.

marksy:
new spacing HOC

- changed name to `withWhiteSpace`
- uses a range of 0-9  margin values (for each small and large screen) based on GDS spacing values
- updated components to be allow recomposing via glamorous

also fixed
- position: duabs (don't know how this got back in there again?)
- widths on unordered/ordered lists
- changed icon colours to black in storybook
- removed floats on radio/checkbox


packages/hoc/src/asWhitespace/index.js
```diff
@@ -0,0 +1,53 @@
...
+  RESPONSIVE_3,
+  RESPONSIVE_4,
+  RESPONSIVE_5,
+  RESPONSIVE_6,
+  RESPONSIVE_7,
+  RESPONSIVE_8,
+  RESPONSIVE_9,
+];
+
+const asWhitespace = (Component, marginBottom) => {
```
penx: maybe `withWhitespace` would be better?

penx: We may add other config options in future so second argument should be a destructured config object `{ marginBottom }` (however see follow up point)

penx:
Step 2 of this is to export every component wrapped in this HOC. That means this PR does not fully resolve #173.

components/button/src/__snapshots__/test.js.snap
```diff
@@ -16,7 +16,6 @@ exports[`button matches wrapper snapshot: wrapper mount 1`] = `
   outline-offset: -1px;
   outline: 1px solid transparent;
   padding: .526315em .789473em .263157em;
-  position: duabs;
```
penx: 👍

PR #188, opened 2018-04-06
updated each component default export as withWhiteSpace HoC

### LINKED ISSUE #173: White space

marksy:
fixes #173 
- changed the default export of each component to utilise the withWhiteSpace HoC
- updated snapshots


components/breadcrumb/src/__snapshots__/test.js.snap
```diff
@@ -1,13 +1,18 @@
...
+}
+
 .glamor-3,
 [data-glamor-3] {
   font-family: nta, Arial, sans-serif;
   font-size: 14px;
   line-height: 16px;
   margin-top: 15px;
-  margin-bottom: 10px;
+  margin-bottom: 0;
```
penx: should the default margin for breadcrumb be 2?

PR #189, opened 2018-04-06
Bugfix/loading box dependency

penx:
`@govuk-react/loading-box` wasn't a sub dependency of govuk-react. We want issues like this to be picked up automatically.

eslint needs to be configured to pick up exports that aren't in package.json, but this can't be done right now, see benmosher/eslint-plugin-import#1049.

In the meantime we are including @govuk-react/storybook as a dependency, which will help ensure the storybook sub dependencies are included.

Also:

- remove package lock, using yarn as standard
- update yarn lock
- remove loading box package lock
- fix to eslint command to include subfolders of src
- add  devDependencies to hoc package
- add  dependencies to storybook package
- govuk-react/storybook as dependency of govuk-react (temporary)


packages/govuk-react/package.json
```diff
@@ -33,6 +33,7 @@
     "@govuk-react/related-items": "^0.1.25",
     "@govuk-react/search-box": "^0.1.25",
     "@govuk-react/select": "^0.1.25",
+    "@govuk-react/storybook": "^0.1.25",
```
stevesims: you haven't added `@govuk-react/loading-box` to this file...

packages/hoc/package.json
```diff
@@ -13,7 +13,35 @@
...
+    "@govuk-react/file-upload": "^0.1.25",
+    "@govuk-react/header": "^0.1.25",
+    "@govuk-react/hint-text": "^0.1.25",
+    "@govuk-react/input": "^0.1.25",
+    "@govuk-react/input-field": "^0.1.25",
+    "@govuk-react/label": "^0.1.25",
+    "@govuk-react/label-text": "^0.1.25",
+    "@govuk-react/list-item": "^0.1.25",
+    "@govuk-react/list-navigation": "^0.1.25",
+    "@govuk-react/multi-choice": "^0.1.25",
```
stevesims: think we should add in `loading-box` here

PR #197, opened 2018-04-10
Feature/Gov UK icon

marksy:
- added GovUK icon
screen shot 2018-04-10 at 12 05 00

[svg source (wikipedia)](https://en.wikipedia.org/wiki/Gov.uk#/media/File:Gov.uk_logo.svg)


penx:
Guess you didn’t mean to add the top nav component?

penx:
We may need to research licensing before we merge this. It may just be a case of setting the license in the specific icon package.json. Or maybe making this a standalone repo with a different license https://en.m.wikipedia.org/wiki/File:Gov.uk_logo.svg#mw-jump-to-license

Issue #198, opened 2018-04-11
If LoadingBox minimum height


penx:
If LoadingBox contains no children and is loading, or the children are less than a certain height, then it should default to a sensible height.

Perhaps this is just setting a min height of 100px ish.

Plos show this in storybook.

marksy:
yes okay

PR #200, opened 2018-04-11
Minimum height for loading box

marksy:
- added min height 100px
- changed default opacity to .85
- showed Storybook example
- updated snapshot

fixex #198

### DIFF
--- a/components/loading-box/src/stories.js
+++ b/components/loading-box/src/stories.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { withKnobs, text, number, boolean } from '@storybook/addon-knobs/react';
-import Header from '@govuk-react/header';
+import Header, { H1, H2 } from '@govuk-react/header';
 import InputField from '@govuk-react/input-field';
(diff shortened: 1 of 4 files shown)


penx:
The opacity change was due to the loader looking odd when covering a lot of bold large text.

Could you please add a story that shows the loader on top of an H1 + H2? Otherwise approved

components/loading-box/src/stories.js
```diff
@@ -119,3 +119,30 @@ stories.add('LoadingBox (long)', () => (
...
+    timeOut={number('timeOut', 200)}
+    backgroundColorOpacity={number('backgroundColorOpacity', 0.85)}
+    spinnerColor={text('spinnerColor', '#000')}
+  >
+    Lorem ipsum dolor sit amet
+  </LoadingBox>
+));
+
+stories.add('with bolded texts', () => (
+  <LoadingBox
```
penx: minor and up to you - I don't think we need knobs for this story, just force loading = true and remove all other props so that they use defaults

marksy: ok

PR #209, opened 2018-04-16
pass text to loading spinner

### LINKED ISSUE #199: Localisation of text in LoadingBox
Text should be localisable, i.e. can be passed in as a prop.

We have text in icons, e.g. the Spinner has 'Loading'.

When a parent component uses an icon, we need to allow this text to be customised by the parent prop.

In the case of the LoadingBox, it should accept a title or similar prop that gets passed down to the spinner icon.

marksy:
fixes #199 

- added label next or below the spinner
- passes label prop to spinner

### DIFF
--- a/components/loading-box/src/index.js
+++ b/components/loading-box/src/index.js
@@ -7,6 +7,12 @@ import { CSSTransition } from 'react-transition-group';
 import hexRgb from 'hex-rgb';
 import { Spinner } from '@govuk-react/icons';
 import { BLACK, WHITE } from 'govuk-colours';
+import {
+  FONT_SIZE,
+  LINE_HEIGHT,
+  MEDIA_QUERIES,
+  NTA_LIGHT,
+} from '@govuk-react/constants';
 
 const spinnerClassName = 'icon-loading';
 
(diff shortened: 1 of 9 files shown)


components/loading-box/src/index.js
```diff
@@ -30,15 +30,19 @@ const Innerwrap = glamorous.div(({
   bottom: 0,
   display: 'flex',
   justifyContent: 'center',
+  alignItems: 'center',
```
penx: is this still needed?

components/loading-box/src/index.js
```diff
@@ -30,15 +30,19 @@ const Innerwrap = glamorous.div(({
...
   [`& .${spinnerClassName}`]: {
-    position: 'absolute',
     zIndex: 101,
     opacity: 1,
     display: 'block',
-    height: '100%',
     maxHeight: 'calc(50vh + 100px)',
     transition: `opacity ${timeIn}ms ease-in-out`,
   },
+  [`& .${spinnerClassName} + p`]: {
```
penx: not sure what all this extra css is needed for just to pass the prop down.. if it's needed could you update the pr description?

marksy:
wasnt breaking lint, just showing as `warn`

PR #211, opened 2018-04-16
Make storybook use src for components rather than lib

### LINKED ISSUE #207: Storybook resolution
We can require the src rather than lib in storybook so that watch isn’t needed

penx:
Make storybook use src for components rather than lib.

This means we don't need to run watch in the background separately, storybook will do this for us.

Fixes #207


packages/storybook/src/stories/index.js
```diff
@@ -1,42 +1,4 @@
...
-
-import '@govuk-react/hoc/lib/asAnchor/stories';
-import '@govuk-react/hoc/lib/asPaginationItem/stories';
-import '@govuk-react/hoc/lib/withWhiteSpace/stories';
-
-import '@govuk-react/icons/lib/ArrowLeft/stories';
-import '@govuk-react/icons/lib/ArrowRight/stories';
-import '@govuk-react/icons/lib/Search/stories';
-import '@govuk-react/icons/lib/Spinner/stories';
+const req = require.context('../../../../', true, /(packages|components)\/[^/]+\/src\/([^/]+\/)*stories.js$/);
```
marksy: crikey

Issue #208, opened 2018-04-14
Use fixtures for storybook and jest snapshot


penx:
Share fixtures file for each component between storybook and jest snapshots 

https://youtu.be/AkCeYUWncjk

penx:
https://github.com/storybooks/storybook/tree/master/addons/storyshots

penx:
Need to consider whether this is compatible with storybook addons such as knobs, or to have some fixtures that are storybook specific stay in storybook

penx:
Mark did some work on this in #213, need to continue to do this in new PRs.

penx:
I think:

- there shouldn't be any fixtures in stories.js or test.js
- all fixtures should be in fixtures.js
- all fixtures should be used by stories.js
- all fixtures should be used by test.js
- example.js should use the default export from fixtures.js and perhaps be deprecated

stevesims:
We've since had second thoughts on this.

In reality, it's not as useful as we'd first thought to share fixtures between tests and stories.  Having fixtures in a separate file makes both tests and stories harder to understand when reviewing their code, as they are by definition not self-contained.  For unit tests at least this is against best practice.

Our create-component script has since been updated and will no longer generate separate a fixture file.

Closing

PR #213, opened 2018-04-17
Feature/table

marksy:
- added a new table component that has styling for table components.
screen shot 2018-04-17 at 14 06 20

### DIFF
--- a/components/table/src/test.js
+++ b/components/table/src/test.js
@@ -1,16 +1,31 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
+import { mount } from 'enzyme';
 import Table from './';
 import Row from './atoms/Row';
 import Cell from './atoms/Cell';
 
-const example = (
-  <Row><Cell>Example</Cell></Row>
+const example = <Table body={<Row><Cell>Example</Cell></Row>} />;
+const exampleWithHead = (
+  <Table
+    body={<Row><Cell>Example</Cell></Row>}
+    head={<Row><Cell>Example</Cell></Row>}
+  />
 );
+
 
 describe(Table, () => {
   it('renders without crashing', () => {
     const div = document.createElement('div');
-    ReactDOM.render(<Table body={example} />, div);
+    ReactDOM.render(example, div);
+    ReactDOM.render(exampleWithHead, div);
+  });
+
+  it('matches snapshot for example', () => {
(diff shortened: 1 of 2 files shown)


components/table/src/index.js
```diff
@@ -0,0 +1,58 @@
...
+);
+
+Table.defaultProps = {
+  caption: undefined,
+  head: undefined,
+};
+
+Table.propTypes = {
+  body: PropTypes.node.isRequired,
+  head: PropTypes.node,
```
penx: What about foot pls? Or is there no concept of a foot in GDS styles?

components/table/src/test.js
```diff
@@ -0,0 +1,16 @@
...
+import Cell from './atoms/Cell';
+
+const example = (
+  <Row><Cell>Example</Cell></Row>
+);
+
+describe(Table, () => {
+  it('renders without crashing', () => {
+    const div = document.createElement('div');
+    ReactDOM.render(<Table body={example} />, div);
```
penx: needs snapshots?

components/table/src/stories.js
```diff
@@ -0,0 +1,69 @@
...
+    <Row>
+      <CellHeader>April</CellHeader>
+      <Cell alignRight>£136.00</Cell>
+      <Cell alignRight>£70.00</Cell>
+    </Row>
+  </React.Fragment>
+);
+
+storiesOf('Table', module).add('default', () => (
+  <Table caption="Dates and amounts" body={example1Body} />
```
penx: See #208 

at some point we need to consider adding a `fixtures.js` file for the stories that is shared by snapshots and storybook? This would also need to be added to our eslint settings in the root. I have a PR open that changes the eslint settings so maybe not a good idea to change them until this is merged. Doesn't need to be done as part of this PR.

marksy: Okay can u plas create an issue/improvement to ‘start using fixtures.js’

penx: I have, it's called #208

marksy:
Thumb 🦃

components/table/src/index.js
marksy: no concept of a foot in GDS

components/table/src/test.js
marksy: okay

marksy:
Okay @penx if you approve i can _merge_

PR #214, opened 2018-04-17
Move devDependencies to root

### LINKED ISSUE #175: Lerna updated still reporting packages that haven't updated
For some reason, when opening PR #174, packages that hadn't been updated were still marked as updated, despite having all tags pulled locally.

See

https://github.com/lerna/lerna/issues/1269
https://github.com/lerna/lerna/issues/1345

penx:
Fixes #175 (hopefully)

Updates to a devDependency for a package causes `lerna publish` to republish that package. This has a knock on effect throughout our repo meaning a single change to one component can result in every component being republished. We don't want this to happen so are including most devDependencies at the project root (where we expect these dependencies are only used when using the package as part of a monorepo).

Exceptions are where a component has specific peerDependencies and we want to also represent that in the devDependencies.

Related: 

lerna/lerna#1269
lerna/lerna#1345
benmosher/eslint-plugin-import#935


components/loading-box/package.json
```diff
@@ -19,22 +19,7 @@
...
-    "@govuk-react/radio": "^0.1.27",
-    "@storybook/addon-actions": "^3.4.1",
-    "@storybook/addon-knobs": "^3.3.15",
-    "@storybook/addons": "^3.4.1",
-    "@storybook/react": "^3.4.1",
-    "enzyme": "^3.3.0",
-    "react-dom": "^16.2.0",
-    "react-transition-group": "^2.2.1",
-    "rimraf": "^2.6.2"
+    "react-transition-group": "^2.2.1"
```
marksy: is this left in devDep bc its the only component that uses this dependency?

penx: yes and it's a peerDependency of the component

PR #226, opened 2018-04-23
Feature/normalise file upload label

Loque-:
Whilst browsing Storybook I noticed that the `FileUpload` component rendered bold label text when there was an error, unlike any of the other form fields.

This change resolves this by removing the `error` prop passed to the `LabelText` component in `FileUpload` bringing it inline with the other form components and stops the label from rendering bold when there is an error.

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged

### DIFF
--- a/components/file-upload/src/index.js
+++ b/components/file-upload/src/index.js
@@ -40,7 +40,7 @@ const FileUpload = ({
   meta, children, hint, acceptedFormats, className,
 }) => (
   <Label className={className} error={meta.error}>
-    <LabelText error={meta.error}>{children}</LabelText>
+    <LabelText>{children}</LabelText>
     {hint && <HintText>{hint}</HintText>}
     {meta.touched && meta.error && <ErrorText>{meta.error}</ErrorText>}
     <GInput type="file" accept={acceptedFormats} error={meta.error} />
--- a/components/file-upload/src/index.js
+++ b/components/file-upload/src/index.js
@@ -15,7 +15,7 @@ import {
   MEDIA_QUERIES,
 } from '@govuk-react/constants';
 
-const Input = glamorous.input({
+const GInput = glamorous.input({
   boxSizing: 'border-box',
   fontFamily: NTA_LIGHT,
   WebkitFontSmoothing: 'antialiased',
@@ -43,7 +43,7 @@ const FileUpload = ({
     <LabelText error={meta.error}>{children}</LabelText>
     {hint && <HintText>{hint}</HintText>}
     {meta.touched && meta.error && <ErrorText>{meta.error}</ErrorText>}
-    <Input type="file" accept={acceptedFormats} error={meta.error} />
+    <GInput type="file" accept={acceptedFormats} error={meta.error} />
   </Label>
 );
 


components/file-upload/src/index.js
```diff
@@ -40,10 +40,10 @@ const FileUpload = ({
   meta, children, hint, acceptedFormats, className,
 }) => (
   <Label className={className} error={meta.error}>
-    <LabelText error={meta.error}>{children}</LabelText>
+    <LabelText>{children}</LabelText>
```
penx: I'm not sure why LabelText has a prop 'error' in the first place, or why it would cause the label to be bold, if this isn't an established pattern.

Is this prop used anywhere else?

For reference: 

https://govuk-elements.herokuapp.com/form-elements/example-form-elements/

components/file-upload/src/index.js
```diff
@@ -15,7 +15,7 @@ import {
   MEDIA_QUERIES,
 } from '@govuk-react/constants';
 
-const Input = glamorous.input({
+const GInput = glamorous.input({
```
Loque-: I spoke to Alasdair about this and he told me this convention had been put in place whenever there were multiple elements with the same name but also required a reference to Glamorous - which means the change I made here is not really required.

components/file-upload/src/index.js
Loque-: Ah yes, I think it was added by mistake, it is also here: https://github.com/penx/govuk-react/blob/master/components/date-input/src/index.js#L77

It appears to make it bold due to this here: https://github.com/penx/govuk-react/blob/master/components/label-text/src/index.js#L33

I've noticed an un-required property on date-input also which I'll add to this branch.

penx: my point is that maybe https://github.com/penx/govuk-react/blob/master/components/label-text/src/index.js#L33 should be removed

Loque-: I agree and from what I can tell it can be removed as Labels should not render bold when there is an error (I cannot see any pattern or suggestion that they should be yet - perhaps this component was used to also render the error text at one point).

penx:
I'm a bit confused as to what the intentional behaviour is, the following screenshot is from  https://govuk-elements.herokuapp.com/form-elements/example-form-elements/ 

screen shot 2018-04-24 at 20 46 03

It shows a drop down with a bold label and regular hint.

The file upload has a regular label and no hint.

I'm not sure on:

1. why the inconsistency in label font weight?
2. is it possible to have a file upload with both a hint and a label?
3. what does govuk-frontend do?

Loque-:
Ah, thanks for that link, I wasn't aware of that.

> 1. why the inconsistency in label font weight?

I am not sure why this is the case, however I think we should support both and can add this in so that it is an option for whoever is implementing it. Would you agree with this?

> 2. is it possible to have a file upload with both a hint and a label?

By looking at the component code and both the `govuk-elements` this appears to be supported. We currently do not have an example of the file upload component without a hint in storybook.

 > 3. what does govuk-frontend do?

They do not support the bold formatting: https://github.com/alphagov/govuk-frontend/blob/master/src/label/_label.scss

components/file-upload/src/index.js
marksy: sorry

marksy:
I think the confusion was because some example fields had bold labels and some didn't. It was like, some had bold when there was an error and  then some didn't

Loque-:
That makes sense - it looks like this was never locked down previously and was left as an option for people implementing a form label to decide whether they wanted it regular or bold.

From the highest priority link you provided Alasdair, their form label implementation allows an optional class: https://github.com/alphagov/govuk-frontend/blob/master/src/label/_label.scss

So I guess I should confirm whether we want to carry this through on this project, and, if everyone else is happy for me to make the change on this PR?

components/file-upload/src/index.js
Loque-: nonsense - thanks so much for all the work on this! Miss you, hope all is good!

PR #234, opened 2018-04-27
Feature/readme clarifications

penx:
readme clarifications


penx:
from @nickcolley:

> GOV.UK Frontend is not an implementation of the GOV.UK Design System, it is a core part of the Design System, so govuk-react would therefore be a port of GOV.UK Frontend.

Will discuss further next week

Issue #238, opened 2018-05-01
viewBox on SVGBase


penx:
Why is the viewBox on SVGBase set to:

+      viewBox="-0.2 0 17 14"

?

Loque-:
That will most likely be a mistake from an authors editor. I'll tidy this up quickly later unless you beat me to it.

Loque-:
I doubt it will be a visible difference to be fair, or if so, just an improvement as I doubt any of the SVG's expected/depended on the negative top value of `-0.2`? Can you think of a good example to check after making the change?

penx:
Looks like only ArrowLeft and ArrowRight use the SVGBase without setting a viewbox, so they're the only 2 to check. Just need to check if the edges of the vector are against the edges before and after.

I also wonder how they would behave without a viewBox set at all.. maybe SVGBase shouldn't set a default viewbox.

Loque-:
I am in favour of removing the default but also ensuring SVG elements have their own viewbox property that is accurate.

For reference: https://www.sarasoueidan.com/blog/svg-coordinate-systems/

PR #250, opened 2018-05-09
Feature/tidy svg

### LINKED ISSUE #243: Width, height and display:block on SVGBase
Should the width and height of the SVGBase be set to 100%?

And display: block set by default?

I have an issue where I'm trying to align an icon that's using SVGBase using `vertical-align: baseline`;

Unfortunately the icon grows to the height of the container, centering the vector in the vertical center, and not allowing me to line up the bottom of the icon with the bottom of the text.

Loque-:
Closes #243, #238 

- Add `viewBox` to SVG assets without any [so they can scale](https://css-tricks.com/scale-svg/)
- Improve Stories with colour, width and height knobs and sensible size defaults
- Remove [un-required](https://stackoverflow.com/a/34249810) and potentially bug-inducing default attributes

* [x] Documentation
* [x] Tests
* [x] Ready to be merged


penx:
Story - "Button: Start with icon" is now broked as icon is not on button. Similar for Search and SearchBox

PR #265, opened 2018-05-14
Remove blank Storybook entry

Loque-:
Un-required storybook entry for SVGBase. I am not sure if it is needed in storybook, or can I think of a good way of representing it in Storybook. Will remove for the moment to avoid any confusion.

* [x] Documentation N/A
* [x] Tests N/A
* [x] Ready to be merged

### DIFF
--- a/packages/icons/src/SVGBase/stories.js
+++ b/packages/icons/src/SVGBase/stories.js
@@ -1,8 +1,0 @@
-import React from 'react';
-import { storiesOf } from '@storybook/react';
-
-import SVG from '.';
-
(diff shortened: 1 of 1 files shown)


penx:
It could be in storybook with a basic circle/rect/vector child

Loque-:
Not sure why Chromatic thinks these are new specs, I created the branch fresh from an updated master with the TopNav already in it.

Loque-:
> It could be in storybook with a basic circle/rect/vector child

Cool, I can add that!

PR #275, opened 2018-05-22
Feature/chromatic auto accept

Loque-:
Could still do with a review to check the conditional comment is ok and meets the conditions expected/required.

(The thing I missed earlier was that `travis.yml` was calling a NPM script and _not_ Chromatic directly, there is also a typo in their documentation `chromatic --auto-accept-changes` should be `chromatic test --auto-accept-changes` which I have raised with them).

Cheers!!

* [x] Documentation
* [x] Tests
* [x] Ready to be merged


Loque-:
I think the last build failed due to NPM connectivity issues.

PR #276, opened 2018-05-22
Feature/paragraph component

### LINKED ISSUE #223: Discussion: how to handle a paragraph/text component?
How would we support a parent application being able to form a block of text that supports `a`, `em` child elements?

Issues to consider:

- If we allow markdown to be passed in as a prop then we'd need to bundle a markdown parser, which would be a large amount of JS
- If we rely on react-dom elements it may not be portable to react-native
- Would a Text/Paragraph component allow other React components as children or should it be a leaf node/escape hatch? If we style HTML elements, they will cascade if the Text or Paragraph component has other React component children. If we don't allow child elements, how do we handle the need for a React Router link?

Loque-:
Adds paragraph component with react-markdown, restricted to only allow Links, Bold and Italics. Whilst GDS advises against Bold and Italics they should still be supported to ensure the component can be used by as many teams as possible.

This PR also contains ESLint fixes (apologies for mixing them into this PR)

closes #223 

* [x] Documentation
* [x] Tests
* [x] Ready to be merged


components/paragraph/src/link-renderer.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Link } from 'react-router-dom';
+import Anchor from './atoms/anchor';
+
+// Support react-router links
+// https://github.com/rexxars/react-markdown/issues/29#issuecomment-231556543
+function LinkRenderer({ href, children }) {
```
stevesims: as a functional component it's probably more conventional for this to be an arrow function :grin:

components/paragraph/src/index.js
```diff
@@ -0,0 +1,81 @@
...
+   */
+  supportingText: PropTypes.bool,
+};
+
+Paragraph.defaultProps = {
+  children: '',
+  supportingText: false,
+};
+
+export default withWhiteSpace({ marginBottom: 5 })(Paragraph);
```
stevesims: isn't the syntax for `withWhiteSpace` to use `mb: 5`?

components/paragraph/src/link-renderer.js
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    href.match(/^\//)
+      ? <Link to={href}>{children}</Link>
+      : <Anchor href={href}>{children}</Anchor>
+  );
+}
+
+LinkRenderer.propTypes = {
+  href: PropTypes.string,
+  children: PropTypes.string,
```
stevesims: `children` should be `PropTypes.node`

components/paragraph/src/test.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { MemoryRouter } from 'react-router-dom';
+import { mount } from 'enzyme';
+
+import Paragraph from './';
+import LinkRenderer from './link-renderer';
```
stevesims: we should consider splitting the `LinkRenderer` tests out to a separate test file

components/paragraph/src/test.js
```diff
@@ -0,0 +1,47 @@
...
+    ReactDOM.render(<Paragraph>Lorem ipsum</Paragraph>, div);
+  });
+
+  it('matches wrapper snapshot', () => {
+    expect(mount(wrapper)).toMatchSnapshot('wrapper mount');
+  });
+});
+
+describe(LinkRenderer, () => {
+  it('returns a react-router `Link` with a relative href path', () => {
```
stevesims: consider adding a test for `#anchor` links

components/paragraph/src/index.js
Loque-: `mb=""` is used as an attribute on JSX components, but when added as a HoC the attribute name appears to be `marginBottom`; https://github.com/penx/govuk-react/blob/master/components/button/src/index.js#L86

components/paragraph/src/link-renderer.js
Loque-: I would rather it stayed as a string as I only want the component to support strings and not nested nodes. AFAIK Markdown should be treated as a string.

components/paragraph/src/test.js
Loque-: Definitely, thanks for this and the ticket!

components/paragraph/src/link-renderer.js
Loque-: Great shout, thank you!

components/paragraph/src/test.js
Loque-: Great shout, thank you, added!

components/paragraph/src/index.js
penx: Perhaps we need an issue to review consistency in future

components/paragraph/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "name": "@govuk-react/paragraph",
+  "version": "0.2.1",
+  "private": false,
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@govuk-react/constants": "^0.2.1",
+    "@govuk-react/hoc": "^0.2.1",
+    "@storybook/addon-knobs": "^3.4.2",
```
penx: Storybook shouldn’t be a production dependency

components/paragraph/src/index.js
stevesims: yeah maybe - the difference is a little bit confusing :grin:

but i am often easily confused... 😕

components/paragraph/package.json
Loque-: I noticed this in some other files, I'll make a note to move it from these to devDependancies also. Cheers for the review!

PR #279, opened 2018-05-23
Feature/remove explicit class name prop

taranchauhan:
**Checklist**:
* [x] Documentation
* [x] Tests
* [x] Ready to be merged 

closes #279

### DIFF
--- a/components/back-link/src/index.js
+++ b/components/back-link/src/index.js
@@ -55,8 +55,8 @@ const Anchor = styled('button')({
   },
 });
 
-const BackLink = ({ children, goBack, ...props }) => (
-  <Anchor onClick={goBack} {...props}>{children}</Anchor>
+const BackLink = ({ goBack, ...props }) => (
+  <Anchor {...props} onClick={goBack} />
 );
 
 BackLink.propTypes = {
--- a/components/back-link/src/index.js
+++ b/components/back-link/src/index.js
@@ -55,8 +55,8 @@ const Anchor = styled('button')({
   },
 });
 
-const BackLink = ({ children, className, goBack }) => (
-  <Anchor className={className} onClick={goBack}>{children}</Anchor>
+const BackLink = ({ children, goBack, ...props }) => (
+  <Anchor onClick={goBack} {...props}>{children}</Anchor>
 );
 
 BackLink.propTypes = {
@@ -68,12 +68,10 @@ BackLink.propTypes = {
    * A function that is called on click
    */
   goBack: PropTypes.func, // TODO: rename onClick
-  className: PropTypes.string,
 };
 
 BackLink.defaultProps = {
   goBack: undefined,
-  className: undefined,
 };
 
 export default withWhiteSpace({ marginBottom: 3 })(BackLink);
--- a/components/checkbox/src/index.js
+++ b/components/checkbox/src/index.js
@@ -88,23 +88,25 @@ const StyledLabel = styled('span')({
 });
 
 const Checkbox = ({
-  children, className, inline, ...input
+  children, inline, disabled, defaultChecked, ...props
 }) => (
-  <StyledCheckbox className={className} inline={inline}>
-    <StyledInput type="checkbox" {...input} />
+  <StyledCheckbox inline={inline} {...props}>
+    <StyledInput type="checkbox" disabled={disabled} defaultChecked={defaultChecked} />
     <StyledLabel>{children}</StyledLabel>
   </StyledCheckbox>
 );
 
 Checkbox.defaultProps = {
-  className: undefined,
   inline: undefined,
+  disabled: false,
+  defaultChecked: false,
 };
 
 Checkbox.propTypes = {
   children: PropTypes.node.isRequired,
-  className: PropTypes.string,
   inline: PropTypes.bool,
+  disabled: PropTypes.bool,
+  defaultChecked: PropTypes.bool,
 };
 
 export default withWhiteSpace({ marginBottom: 2 })(Checkbox);
--- a/components/checkbox/src/index.js
+++ b/components/checkbox/src/index.js
@@ -88,25 +88,23 @@ const StyledLabel = styled('span')({
 });
 
 const Checkbox = ({
-  children, inline, disabled, defaultChecked, ...props
+  children, className, inline, ...props
 }) => (
-  <StyledCheckbox inline={inline} {...props}>
-    <StyledInput type="checkbox" disabled={disabled} defaultChecked={defaultChecked} />
+  <StyledCheckbox inline={inline} className={className}>
+    <StyledInput type="checkbox" {...props} />
     <StyledLabel>{children}</StyledLabel>
   </StyledCheckbox>
 );
 
 Checkbox.defaultProps = {
+  className: undefined,
   inline: undefined,
-  disabled: false,
-  defaultChecked: false,
 };
 
 Checkbox.propTypes = {
   children: PropTypes.node.isRequired,
+  className: PropTypes.string,
   inline: PropTypes.bool,
-  disabled: PropTypes.bool,
-  defaultChecked: PropTypes.bool,
 };
 
 export default withWhiteSpace({ marginBottom: 2 })(Checkbox);
--- a/components/file-upload/src/index.js
+++ b/components/file-upload/src/index.js
@@ -36,7 +36,7 @@ const StyledInput = styled('input')({
 const FileUpload = ({
   meta, children, hint, acceptedFormats, ...props
 }) => (
-  <Label error={meta.error} {...props}>
+  <Label {...props} error={meta.touched && meta.error}>
     <LabelText error={meta.error}>{children}</LabelText>
     {hint && <HintText>{hint}</HintText>}
     {meta.touched && meta.error && <ErrorText>{meta.error}</ErrorText>}
--- a/components/file-upload/src/index.js
+++ b/components/file-upload/src/index.js
@@ -34,9 +34,9 @@ const StyledInput = styled('input')({
 });
 
 const FileUpload = ({
-  meta, children, hint, acceptedFormats, className,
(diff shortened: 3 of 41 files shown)


components/back-link/src/index.js
```diff
@@ -55,8 +55,8 @@ const Anchor = styled('button')({
   },
 });
 
-const BackLink = ({ children, className, goBack }) => (
-  <Anchor className={className} onClick={goBack}>{children}</Anchor>
+const BackLink = ({ children, goBack, ...props }) => (
+  <Anchor onClick={goBack} {...props}>{children}</Anchor>
```
stevesims: was wondering if we should place these things around the other way, i.e. `<Anchor {...props} onClick={goBack}>` but doing it this way around means that we can support using `onClick` as a prop which would override `goBack`, which fits nicely with the notion that we should rename `goBack` to `onClick` - i.e. we're already supporting that :grin:

components/checkbox/src/index.js
```diff
@@ -88,23 +88,25 @@ const StyledLabel = styled('span')({
 });
 
 const Checkbox = ({
-  children, className, inline, ...input
+  children, inline, disabled, defaultChecked, ...props
```
stevesims: i'm not so sure about these changes

the intent of the `checkbox` component is that it works the same as `<input type="checkbox">`...

components/checkbox/src/index.js
```diff
@@ -88,23 +88,25 @@ const StyledLabel = styled('span')({
...
 }) => (
-  <StyledCheckbox className={className} inline={inline}>
-    <StyledInput type="checkbox" {...input} />
+  <StyledCheckbox inline={inline} {...props}>
```
stevesims: i don't think that we should have {...props} here but instead...

components/checkbox/src/index.js
```diff
@@ -88,23 +88,25 @@ const StyledLabel = styled('span')({
...
+    <StyledInput type="checkbox" disabled={disabled} defaultChecked={defaultChecked} />
```
stevesims: this is where i think we should have `{...props}`

which means that you wouldn't need to explicitly extract `disabled` and `defaultChecked`

components/checkbox/src/index.js
stevesims: however this may well be the place that we want to put any `className` that may have been passed in...

components/file-upload/src/index.js
```diff
@@ -34,9 +34,9 @@ const StyledInput = styled('input')({
 });
 
 const FileUpload = ({
-  meta, children, hint, acceptedFormats, className,
+  meta, children, hint, acceptedFormats, ...props
 }) => (
-  <Label className={className} error={meta.error}>
+  <Label error={meta.error} {...props}>
```
stevesims: we might want to have this as `<Label {...props} error={meta.error}>` so as to ensure that we don't over-write the `error` prop we're passing through here with another `error` value passed into this component

stevesims: also think that this should probably be `error={meta.touched && meta.error}`

components/input-field/src/index.js
```diff
@@ -11,9 +11,9 @@ import Input from '@govuk-react/input';
 import { withWhiteSpace } from '@govuk-react/hoc';
 
 const InputField = ({
-  meta, children, hint, input, className,
+  meta, children, hint, input, ...props
 }) => (
-  <Label className={className} error={meta.touched && meta.error}>
+  <Label error={meta.touched && meta.error} {...props}>
```
stevesims: as per previous suggestion, we probably want to spread `props` first

components/layout/src/index.js
```diff
@@ -17,7 +17,7 @@ const StyledLayout = styled('div')({
   },
 });
 
-const Layout = ({ children }) => <StyledLayout>{children}</StyledLayout>;
+const Layout = ({ children, ...props }) => <StyledLayout {...props}>{children}</StyledLayout>;
```
stevesims: this can be just `const Layout = props => <StyledLayout {...props} />`
as `children` is a prop

components/lead-paragraph/src/index.js
```diff
@@ -23,8 +23,8 @@ const StyledParagraph = styled('p')({
   },
 });
 
-const LeadParagraph = ({ children }) => (
-  <StyledParagraph>{children}</StyledParagraph>
+const LeadParagraph = ({ children, ...props }) => (
```
stevesims: this can be `const LeadParagraph = props => <StyledParagraph {...props} />`

components/list-item/src/index.js
```diff
@@ -24,17 +24,12 @@ const StyledListItem = styled('li')({
   },
 });
 
-const ListItem = ({ children, className }) => (
-  <StyledListItem className={className}>{children}</StyledListItem>
+const ListItem = ({ children, ...props }) => (
```
stevesims: can also be simplified

components/loading-box/src/index.js
```diff
@@ -130,10 +130,9 @@ const LoadingBox = ({
   spinnerColor,
   timeIn,
   timeOut,
-  ...props
 }) => (
   <StyledContainer>
```
stevesims: maybe we should be spreading `prop` here?

components/pagination/src/index.js
```diff
@@ -33,17 +33,12 @@ const StyledList = styled('ul')({
   },
 });
 
-const Pagination = ({ children, className }) => (
-  <StyledList className={className}>{children}</StyledList>
+const Pagination = ({ children, ...props }) => (
```
stevesims: this can be simplified

components/radio/src/index.js
```diff
@@ -84,25 +84,39 @@ const LabelText = styled('span')({
 });
 
 const Radio = ({
-  inline, children, className, disabled, ...input
+  inline, children, disabled, checked, name, value, input, ...props
```
stevesims: i'm not sure about these changes

this component is now expecting an `input` object in it's props for values to pass through to the `input` element

it's probably wiser to keep this component as it was

components/related-items/src/index.js
```diff
@@ -32,17 +32,12 @@ const StyledRelatedItems = styled('div')({
   },
 });
 
-const RelatedItems = ({ children, className }) => (
-  <StyledRelatedItems className={className}>{children}</StyledRelatedItems>
+const RelatedItems = ({ children, ...props }) => (
```
stevesims: this can be simplified

components/select/src/index.js
```diff
@@ -44,9 +44,9 @@ const StyledSelect = styled('select')(
 );
 
 const Select = ({
-  children, hint, label, meta, input, className,
+  children, hint, label, meta, input, ...props
 }) => (
-  <Label className={className} error={meta.touched && meta.error}>
+  <Label error={meta.touched && meta.error} {...props}>
```
stevesims: probably should swap these props around too

components/text-area/src/index.js
```diff
@@ -43,31 +43,31 @@ const TextAreaField = styled('textarea')(
...
-    {props.hint && <HintText>{props.hint}</HintText>}
-    {props.meta.touched &&
-      props.meta.error && <ErrorText>{props.meta.error}</ErrorText>}
+const TextArea = ({
+  children, hint, meta, ...input
+}) => (
+  <Label error={meta.touched && meta.error}>
+    <LabelText>{children}</LabelText>
+    {hint && <HintText>{hint}</HintText>}
+    {meta.touched &&
```
stevesims: this could go onto one line
(unless the linting complains?)

components/unordered-list/src/index.js
```diff
@@ -38,19 +38,17 @@ const StyledList = styled('ul')(
 
 // listStyleType: normal HTML property values are used here
 // e.g., listStyleType="circle", listStyleType="upper-alpha", listStyleType="none"
-const UnorderedList = ({ children, listStyleType, className }) => (
-  <StyledList className={className} listStyleType={listStyleType}>{children}</StyledList>
+const UnorderedList = ({ children, listStyleType, ...props }) => (
+  <StyledList listStyleType={listStyleType} {...props}>{children}</StyledList>
```
stevesims: this can be simplified, removing explicit `children`

components/text-area/src/index.js
```diff
@@ -43,31 +43,25 @@ const TextAreaField = styled('textarea')(
...
-    {props.meta.touched &&
-      props.meta.error && <ErrorText>{props.meta.error}</ErrorText>}
-    <TextAreaField
-      type="text"
-      rows="5"
-      error={props.meta.touched && props.meta.error}
-      {...props.input}
-    />
+const TextArea = ({
+  children, hint, meta, ...input
```
stevesims: you need to add in `input` here as an explicit prop

as you can see above in the previous code, `TextAreaField` was getting `{...prop.input}`

so this should be `children, hint, meta, input, ...props`

components/text-area/src/index.js
```diff
@@ -43,31 +43,25 @@ const TextAreaField = styled('textarea')(
...
+}) => (
+  <Label error={meta.touched && meta.error}>
```
stevesims: this should gain `{...props}`

components/text-area/src/index.js
```diff
@@ -43,31 +43,25 @@ const TextAreaField = styled('textarea')(
...
+    <LabelText>{children}</LabelText>
+    { hint && <HintText>{hint}</HintText> }
+    { meta.touched && meta.error && <ErrorText>{meta.error}</ErrorText> }
+    <TextAreaField type="text" rows="5" error={meta.touched && meta.error} {...input} />
```
stevesims: this stays the same :grin:

components/text-area/src/index.js
taranchauhan: Oops, good spot! Completely missed the .input at the end of ...props when making this change

PR #285, opened 2018-05-24
Fix button styling issues:

### LINKED ISSUE #280: Button component  is not accurate to GDS element styling
**Button component needs to add the following**:

Hover state:
- Colour should darken slightly

Active state:
- Add Top 2px
- Remove box shadow

See [GDS Buttons](https://govuk-elements.herokuapp.com/buttons/#)

taranchauhan:
Hover state:
Background colour should use GDS button hover state

Active state:
Background colour should use GDS button hover state
Add Top 2px
Remove box shadow

**Checklist**:
* [x] Documentation
* [x] Tests
* [x] Ready to be merged 

closes #280

### DIFF
--- a/components/button/src/index.js
+++ b/components/button/src/index.js
@@ -43,11 +43,11 @@ const StyledButton = styled('button')(
       backgroundColor: GREEN,
     },
     ':focus': {
-      position: 'absolute',
+      position: 'relative',
       top: '2px',
       backgroundColor: GREEN,
       outline: `3px solid ${YELLOW}`,
-      boxShadow: 'none',
+      boxShadow: '0 0px 0',
     },
     // TODO: avoid cascade
     ' svg': {
--- a/components/button/src/index.js
+++ b/components/button/src/index.js
@@ -13,6 +13,7 @@ import { NTA_LIGHT, SPACING } from '@govuk-react/constants';
 import {
   BUTTON_COLOUR,
   BUTTON_COLOUR_DARKEN_15,
+  GREEN,
   WHITE,
   YELLOW,
 } from 'govuk-colours';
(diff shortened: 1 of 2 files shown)


components/button/src/index.js
```diff
@@ -38,8 +39,15 @@ const StyledButton = styled('button')(
     textDecoration: 'none',
     WebkitAppearance: 'none',
     WebkitFontSmoothing: 'antialiased',
+    ':hover': {
+      backgroundColor: GREEN,
+    },
     ':focus': {
+      position: 'absolute',
```
Loque-: `relative` would be more suitable here

components/button/src/index.js
```diff
@@ -38,8 +39,15 @@ const StyledButton = styled('button')(
...
+      top: '2px',
+      backgroundColor: GREEN,
       outline: `3px solid ${YELLOW}`,
+      boxShadow: 'none',
```
Loque-: Instead of setting it to `none`, I think it could be better for browsers to adjust the value to `0px` (could have a potential performance impact).

stevesims:
could be good to link to the GDS example for this

components/button/src/index.js
```diff
@@ -38,8 +39,15 @@ const StyledButton = styled('button')(
     textDecoration: 'none',
     WebkitAppearance: 'none',
     WebkitFontSmoothing: 'antialiased',
+    ':hover': {
```
stevesims: looking around a bit wider at the rest of the styling for button here, i'm not sure how it matches up with GDS

the example here https://govuk-elements.herokuapp.com/buttons/ seems to use a slightly different approach especially WRT how `top` works.  the `:focus` state for example doesn't adjust `top`, however `:active` does

PR #290, opened 2018-05-25
Feature/top nav styling 283

### LINKED ISSUE #283: TopNav component is not accurate to GDS element styling
- Top-nav vertical padding from 10px to 8px
- Visited top-nav colour
- Pseudo states top-nav (hover and active)
 - Underline property (text-decoration-skip-ink)

taranchauhan:
**CHANGELOG**:
- Update TopNav vertical padding from 10px to 8px
- Create LogoAnchor atom that takes in an anchorType 'a' or Link (Required due to different styling of GOV.UK Logo)
- Create NavLinkAnchor atom that takes in an anchorType 'a' or Link (Required due to different styling of NavLink's in TopNav)
- Apply LogoAnchor and NavLink anchor pseudo states for hover and focus to match GDS elements
- Storybook examples using LogoAnchor and NavLinkAnchor with React Router <Link>
- Create Tests for LogoAnchor and NavLinkAnchor (Snapshot and prop tests)
- Improve existing asAnchor tests (Add prop tests)

**Checklist**:
* [x] Documentation
* [x] Tests
* [x] Ready to be merged

closes #283

### DIFF
--- a/components/top-nav/src/asAnchor.js
+++ b/components/top-nav/src/asAnchor.js
@@ -1,7 +1,5 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import styled from 'react-emotion';
-import { WHITE, YELLOW } from 'govuk-colours';
 
 const asAnchor = (AnchorType) => {
   const Anchor = props => (
@@ -13,26 +11,7 @@ const asAnchor = (AnchorType) => {
       .isRequired,
   };
 
-  const StyledHoc = styled(Anchor)({
-    color: WHITE, // TODO: active state, LIGHT_BLUE
-    textDecoration: 'none',
-    borderBottom: '1px solid transparent',
-    fontWeight: 700,
-    lineHeight: 1,
-    ':hover': {
-      borderBottomColor: WHITE, // TODO: active state, LIGHT_BLUE
-    },
-    ':focus': {
-      backgroundColor: YELLOW,
-      outline: `3px solid ${YELLOW}`,
-    },
-    ':visited': {
-      color: WHITE, // TODO: active state, LIGHT_BLUE
-      borderBottomColor: WHITE, // TODO: active state, LIGHT_BLUE
-    },
-  });
-
-  return StyledHoc;
+  return Anchor;
 };
 
 export default asAnchor;
--- a/components/top-nav/src/atoms/logo-anchor/index.js
+++ b/components/top-nav/src/atoms/logo-anchor/index.js
@@ -32,7 +32,7 @@ LogoAnchor.defaultProps = {
 };
 
 LogoAnchor.propTypes = {
-  anchorType: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+  anchorType: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
 };
 
 export default LogoAnchor;
--- a/components/top-nav/src/atoms/logo-anchor/index.js
+++ b/components/top-nav/src/atoms/logo-anchor/index.js
@@ -1,0 +1,38 @@
+import styled from 'react-emotion';
+import React from 'react';
+import PropTypes from 'prop-types';
+import { WHITE, YELLOW, BLACK } from 'govuk-colours';
+
+const LogoAnchor = ({ anchorType, ...props }) => {
+  const StyledLogoAnchor = styled(anchorType)({
+    color: WHITE, // TODO: active state, LIGHT_BLUE
+    textDecoration: 'none',
+    textDecorationSkipInk: 'none',
+    borderBottom: '1px solid transparent',
+    fontWeight: 700,
+    lineHeight: 1,
+    ':hover': {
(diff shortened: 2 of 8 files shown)


components/top-nav/src/asAnchor.js
```diff
@@ -13,26 +11,7 @@ const asAnchor = (AnchorType) => {
       .isRequired,
   };
 
-  const StyledHoc = styled(Anchor)({
```
stevesims: OK - so - I must admit that when I first looked at this I got a bit confused.  That's mostly because I was thinking of the `asAnchor` HOC that's in the HOC package.  The fact that this HOC was given an identical name wasn't a great choice.  :grin:  Maybe it should have been `asTopNavAnchor`

The idea of having an HOC like this is to avoid code duplication.  Looking through the rest of the code we do now have some duplication between the two anchor atom components.  It would probably be best to have the common styles provided by this HOC, and then only add/override to get the different custom elements of styling as necessary in those individual components.

As it stands right now this file has become a null component that just passes through all the props - it's used only in the unit test file and *could* therefore be deleted.

IMHO it would be best to restore styling into this component and simplify the atoms to remove duplication.

components/top-nav/src/atoms/logo-anchor/index.js
```diff
@@ -0,0 +1,38 @@
+import styled from 'react-emotion';
+import React from 'react';
+import PropTypes from 'prop-types';
+import { WHITE, YELLOW, BLACK } from 'govuk-colours';
+
+const LogoAnchor = ({ anchorType, ...props }) => {
+  const StyledLogoAnchor = styled(anchorType)({
```
stevesims: I like passing in the `anchorType` here 👍 
Seems like a good and sensible pattern.

components/top-nav/src/atoms/top-nav-inner/index.js
```diff
@@ -9,10 +9,10 @@ const TopNavInner = styled('div')({
   flexDirection: 'column',
   width: `calc(100% - ${SPACING.SCALE_3})`,
   maxWidth: '960px',
-  padding: `${SPACING.SCALE_2} ${SPACING.SCALE_3}`,
+  padding: `8px ${SPACING.SCALE_3}`,
```
stevesims: not sure about the magic number here of `8px` - it feels like this should be a variable, given the other use of variables for spacing

components/top-nav/src/atoms/top-nav-inner/index.js
```diff
@@ -9,10 +9,10 @@ const TopNavInner = styled('div')({
   flexDirection: 'column',
   width: `calc(100% - ${SPACING.SCALE_3})`,
   maxWidth: '960px',
```
stevesims: should we get rid of this?

components/top-nav/src/stories.js
```diff
@@ -1,55 +1,82 @@
...
 
 const ServiceTitle = (
-  <AnchorTag color="#fff" href={link} target="new">
-    <React.Fragment>
-      <Header level={3}>Service Title</Header>
-    </React.Fragment>
-  </AnchorTag>
+  <NavLinkAnchor href={link} target="new">
+    <Header level={3}>
+        Service Title
```
stevesims: there's extra spacing that shouldn't be here :grin:

components/top-nav/src/atoms/logo-anchor/index.js
```diff
@@ -0,0 +1,38 @@
...
+
+  return <StyledLogoAnchor {...props} />;
+};
+
+LogoAnchor.defaultProps = {
+  anchorType: 'a',
+};
+
+LogoAnchor.propTypes = {
+  anchorType: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
```
stevesims: I'm wondering if this is entirely correct.  Whether this should be just `PropTypes.node`?

As it stands this definition doesn't allow for class-based components to be used as an `anchorType`

components/top-nav/src/atoms/logo-anchor/index.js
stevesims: As noted against the `asAnchor` HOC source above though, it'd be good to see the style duplication removed from here

components/top-nav/src/atoms/nav-link-anchor/index.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
```
stevesims: same comments apply to this component as for the one above 😀

components/top-nav/src/asAnchor.js
penx: > asTopNavAnchor

yep

components/top-nav/src/atoms/top-nav-inner/index.js
taranchauhan: This allows flex column stacking on small screen sizes.

PR #296, opened 2018-05-30
Feature/inset component

taranchauhan:
**Checklist**:
* [x] Documentation
* [x] Tests
* [x] Ready to be merged

Created <InsetText> component

### DIFF
--- a/components/inset-text/src/fixtures.js
+++ b/components/inset-text/src/fixtures.js
@@ -1,0 +1,9 @@
+import React from 'react';
+import { boolean } from '@storybook/addon-knobs/react';
+import InsetText from '.';
+
+const insetText = 'It can take up to 8 weeks to register a lasting power of attorney if there are no mistakes in the application.';
+
+export default () => <InsetText>{insetText}</InsetText>;
+
+export const InsetTextNarrowBorder = () => <InsetText isNarrow={boolean('IsNarrow', true)}>{insetText}</InsetText>;
--- a/components/inset-text/src/index.js
+++ b/components/inset-text/src/index.js
@@ -10,7 +10,7 @@ import styled from 'react-emotion';
 
 import Paragraph from '@govuk-react/paragraph';
 
-import { SPACING_MAP, SPACING } from '@govuk-react/constants';
+import { SPACING } from '@govuk-react/constants';
 import { BORDER_COLOUR } from 'govuk-colours';
 
 const StyledInsetText = styled(Paragraph)(props => ({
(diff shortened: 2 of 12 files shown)


components/inset-text/src/fixtures.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import { boolean } from '@storybook/addon-knobs/react';
+import InsetText from '.';
+
+const insetText = 'It can take up to 8 weeks to register a lasting power of attorney if there are no mistakes in the application.';
+
+export default () => <InsetText>{insetText}</InsetText>;
```
Loque-: Insert text as knob?

components/inset-text/src/index.js
```diff
@@ -0,0 +1,30 @@
...
+  padding: SPACING.SCALE_3,
+}));
+
+const InsetText = props => <StyledInsetText {...props} mb="3" />;
+
+InsetText.propTypes = {
+  children: PropTypes.node.isRequired,
+};
+
+export default InsetText;
```
Loque-: If you could include that comment about why this component is not exported with whitespace it might be handy for the next person who comes across this.

components/inset-text/src/test.js
```diff
@@ -0,0 +1,38 @@
...
+  it('should render some inset text with a narrow border', () => {
+    const output = shallow(<InsetTextNarrowBorder />);
+    expect(output.find('p')).toBeTruthy();
+  });
+
+  it('matches wrapperInsetTextNarrowBorder snapshot', () => {
+    expect(wrapperInsetTextNarrowBorder).toMatchSnapshot('inset text narrow border');
+  });
+
+  it('should have a narrow border left width ', () => {
```
Loque-: I would recommend adjusting this to `Renders component with isNarrow attribute` as the test does not check the border left width

components/inset-text/src/test.js
```diff
@@ -0,0 +1,38 @@
...
+  it('should render some inset text', () => {
+    const output = shallow(<InsetText />);
+    expect(output.find('p')).toBeTruthy();
+  });
+
+  it('matches wrapperInsetText snapshot', () => {
+    expect(wrapperInsetText).toMatchSnapshot('wrapperInsetText mount');
+  });
+
+  it('should render some inset text with a narrow border', () => {
```
Loque-: I would adjust the name of this test to be something like `Should render a paragraph element when passed text`. I may be being too particular about the wording of these tests however.

Loque-:
Awesome, nice one!

PR #324, opened 2018-06-11
Bugfix/HiddenText should be closed by default

Joe-Brady:
**Checklist**:

* [ ] Documentation N/A
* [x] Tests
* [x] Ready to be merged 

New test fails pre-change:
screen shot 2018-06-11 at 11 43 33

All tests pass post-change:
screen shot 2018-06-11 at 11 44 07

### DIFF
--- a/components/hidden-text/src/test.js
+++ b/components/hidden-text/src/test.js
@@ -32,6 +32,13 @@ describe('hidden text', () => {
     });
   });
 
+  it('should be closed by default', () => {
+    const output = mount(<HiddenText />);
+    const detailsElement = output.find('details');
+
+    expect(detailsElement.get(0).props.open).toEqual(undefined);
+  });
+
   it('matches the HiddenText snapshot', () => {
     expect(wrapperHiddenText).toMatchSnapshot('hidden text');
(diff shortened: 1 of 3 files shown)


components/hidden-text/src/test.js
```diff
@@ -32,6 +32,13 @@ describe('hidden text', () => {
     });
   });
 
+  it('should be closed by default', () => {
+    const output = mount(<HiddenText />);
```
Loque-: You can use shallow here instead, it's a little more performant and people seem to harp on about it being best practice, I did some tests locally as I am pretty new to enzyme but this works nicely;

```
it('should be closed by default', () => {
  const output = shallow(<HiddenText />);
  expect(output.prop('open')).toEqual(undefined);
});
```

Joe-Brady: @Loque- as discussed, discovered that shallow is not sufficient to access the HTML element with the "open" attribute.

Is this now ok to merge?

Loque-: Thanks for the looking at that Joe!

PR #326, opened 2018-06-11
Feature/error summary

taranchauhan:
**Checklist**:
* [x] Documentation
* [x] Tests
* [x] Ready to be merged

- Created ErrorSummary component as referenced in [Error Summary](https://govuk-elements.herokuapp.com/errors/#summarise-errors)
- with tests
- with stories
- with documentation


components/error-summary/src/index.js
```diff
@@ -0,0 +1,153 @@
...
+  </StyledErrorSummary>
+);
+
+ErrorSummary.defaultProps = {
+  description: undefined,
+  errors: [],
+};
+
+ErrorSummary.propTypes = {
+  heading: PropTypes.string.isRequired,
```
Loque-: docs for props needed : )

components/error-summary/src/fixtures.js
```diff
@@ -0,0 +1,41 @@
...
+    targetName: 'national-insurance-number',
+    text: 'National Insurance number error',
+  },
+  {
+    targetName: 'description',
+    text: 'Description of what you saw error',
+  },
+];
+
+export default () => (
```
Loque-: Not sure if it is worth `default` being the component by itself, so it can be tested and rendered in isolation without anything else? Altho, I like the tests, and I can't see any issue with this.

Loque-: Guess it might just my semantic radar getting mixed up, as it is exporting `default`, not that, the thing being exported is the components actual default... if you know what I mean? o_O 

Probably fine to leave as is

Loque-:
Slick!

PR #329, opened 2018-06-13
ErrorSummary: Test onClick for each error in the list of errors by us…

taranchauhan:
**Checklist**:
* [x] Documentation
* [x] Tests
* [x] Ready to be merged

# ErrorSummary

**Problem**
- The onClick for each error was not being tested which brought the coverage below 100%

**Solution**
- Test onClick for each error in the list of errors by passing in onHandleErrorClick as a prop instead.
- A mock jest.fn() is passed in during the test so that the number of invocations of the click can be tested.

### DIFF
--- a/components/error-summary/src/fixtures.js
+++ b/components/error-summary/src/fixtures.js
@@ -1,3 +1,5 @@
+/* eslint-env browser */
+
 import React from 'react';
 import { text } from '@storybook/addon-knobs/react';
 import InputField from '@govuk-react/input-field';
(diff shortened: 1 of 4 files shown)


components/error-summary/src/fixtures.js
```diff
@@ -18,11 +20,16 @@ export const errors = [
   },
 ];
 
+const onHandleErrorClick = (targetName) => {
```
Loque-: Don't know why I didn't think of this, way better, nice one! I think we could iterate on this in the future if we wanted to also... althoooooo context api : D

PR #337, opened 2018-06-19
Fix console error for arrays passed to panel body

anishamalde:
**Checklist**:

* [ ] Documentation n/a
* [x] Tests
* [x] Ready to be merged 

screen shot 2018-06-19 at 16 05 04
screen shot 2018-06-19 at 15 35 36

### DIFF
--- a/components/panel/src/index.js
+++ b/components/panel/src/index.js
@@ -67,13 +67,18 @@ const StyledBody = styled('div')({
  * - https://github.com/alphagov/govuk-frontend/tree/master/src/components/panel
  *
  */
+
+const generatePanelBodyKey = () => {
+  return `panelBody_${Math.floor(Math.random() * Math.floor(1000000))}`;
+}
+
 const Panel = ({ panelTitle, panelBody, ...props }) => (
   <StyledPanel {...props}>
     <StyledTitle>{panelTitle}</StyledTitle>
     <StyledBody>
       {Array.isArray(panelBody)
-        ? panelBody.map((element, index) => (
-          <Fragment key={index}>{element}</Fragment>
+        ? panelBody.map((element) => (
+          <Fragment key={generatePanelBodyKey()}>{element}</Fragment>
         ))
         : panelBody}
     </StyledBody>
@@ -87,7 +92,7 @@ Panel.propTypes = {
   /** Panel title text */
   panelTitle: PropTypes.string.isRequired,
   /** Panel body text */
-  panelBody: PropTypes.oneOfType([PropTypes.string, PropTypes.array])
+  panelBody: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
 };
 
 export default withWhiteSpace({ marginBottom: 3 })(Panel);
--- a/components/panel/src/index.js
+++ b/components/panel/src/index.js
@@ -9,41 +9,41 @@ import {
   SPACING,
 } from '@govuk-react/constants';
 
-const StyledPanel = styled('div')({
+const StyledPanel = styled("div")({
   backgroundColor: TURQUOISE,
   color: WHITE,
-  padding: SPACING.SCALE_5,
+  padding: SPACING.SCALE_5
 });
 
 // TODO use standard font constants
-const StyledTitle = styled('h2')({
+const StyledTitle = styled("h2")({
   fontFamily: NTA_LIGHT,
-  fontWeight: 'bold',
-  WebkitFontSmoothing: 'antialiased',
-  fontSize: '32px',
-  lineHeight: '35px',
+  fontWeight: "bold",
+  WebkitFontSmoothing: "antialiased",
+  fontSize: "32px",
+  lineHeight: "35px",
   marginTop: 0,
-  textAlign: 'center',
+  textAlign: "center",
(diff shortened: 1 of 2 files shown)


Loque-:
Thanks for the PR!

I think you have `prettier` or something running as it has completely ignored the ESLint rules we have in place tho : D, you can see it has reformatted the file and used double quotes instead of single quotes, and removed trailing commas etc.... which is not the change you are trying to merge : )

I know this as I also use prettier on personal projects and Gavin who recently joined our team also enjoys using it, and, you can still use it, but, you will need to configure it ever so slightly for the project. Could I leave this with you to do? If you are using VSCode I might have a suggestion for adding rules for this workspace as I did it on the weekend for a personal project.

Cheers!

components/panel/src/index.js
```diff
@@ -70,18 +70,24 @@ const StyledBody = styled('div')({
 const Panel = ({ panelTitle, panelBody, ...props }) => (
   <StyledPanel {...props}>
     <StyledTitle>{panelTitle}</StyledTitle>
-    <StyledBody>{panelBody}</StyledBody>
+    <StyledBody>
+      {Array.isArray(panelBody)
+        ? panelBody.map((element, index) => (
+          <React.Fragment key={index}>{element}</React.Fragment>
```
Loque-: This bit looks cool tho! It might me more elegant to `import React, { Fragment } from 'React'` at the top of the file instead and then use `Fragment` instead of `React.Fragment`

Joe-Brady:
@Loque- changed as requested! Looks like just the actual changes now. Also added {Fragment}

Loque-:
if you run `yarn build` and then `yarn test` you'll get the ESlint feedback locally btw! I _really_ need to fix these other tests up so they are less noisy and warning-ey : )))

Cheers!

Loque-:
Hey just a couple of lint errors returned by Travis, if you can fix these I can merge this in and include it in the next release;
```
 /home/travis/build/UKHomeOffice/govuk-react/components/panel/src/index.js
  76:26  error  Do not use Array index in keys                 react/no-array-index-key
  90:70  error  Missing trailing comma                         comma-dangle
  93:59  error  Newline required at end of file but not found  eol-last
/home/travis/build/UKHomeOffice/govuk-react/components/panel/src/test.js
  40:30  error  Expected parentheses around arrow function argument having a body with curly braces  arrow-parens
  49:4   error  Newline required at end of file but not found     
```

Joe-Brady:
@Loque- after some research, and due to the linting rule not accepting the array index as a key (but the key being required by React) - this is the solution I came up with - please let me know if that isn't a solution that you're happy with!

```
const generatePanelBodyKey = (pre) => {
  return `panelBody_${Math.floor(Math.random() * Math.floor(1000000))}`;
}

const Panel = ({ panelTitle, panelBody, ...props }) => (
  <StyledPanel {...props}>
    <StyledTitle>{panelTitle}</StyledTitle>
    <StyledBody>
      {Array.isArray(panelBody)
        ? panelBody.map((element) => (
          <Fragment key={generatePanelBodyKey()}>{element}</Fragment>
        ))
        : panelBody}
    </StyledBody>
  </StyledPanel>
);
```

Loque-:
I'm not sure that's necessary, but I can't remember what the solutions are for this, I think @stevesims has a quick answer for overcoming this tho which would be handy for us both as I know I am going to run into this (there might examples already in the codebase somewhere also).

Loque-:
I think it is oki to disable the lint rule for that line based off the documentation for the lint rule; https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md#when-not-to-use-it

Loque-:
In the interest of combined knowledge, @taranchauhan got around it by creating a custom string with the value of index + 1: https://github.com/UKHomeOffice/govuk-react/blob/582f3cd3495c5176029d49d3ee8d933f9ac0821e/components/hidden-text/src/fixtures.js

I personally think it's fine to disable the rule for these use cases tho. I'll find the exclude line now to help.

Loque-:
Ah, hah! Already done here; https://github.com/UKHomeOffice/govuk-react/blob/6cc405020deb0ad382f24296be60fea92fe7f79e/packages/icons/src/Spinner/index.js#L30

gavinorland:
Hey guys, I know what it's like dealing with ESLint and its key complaints. I thought I would mention [this article](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318) which details the problem with using indices. Also the [UUID package](https://www.npmjs.com/package/uuid) - I've seen this used quite often, and used it myself, to generate unique keys on the fly - can be handy!

Loque-:
@gavinorland I don't think this is necessary in this case. I might be wrong, but I'd rather we added the ignore for this instance and avoid adding an additional dependancy. Also there can be issues caused with server-side rendering and UUID generation.

Loque-:
Thanks for the article too btw, really really great to read up on it and discuss. The exceptions to the rule are particularly relevant to our cases for the moment however and I think re-enforce the case of adding the ESLint exception in this case. <3

Joe-Brady:
@Loque- pushed that change to ignore the rule!

PR #347, opened 2018-06-26
Feature/hidden text margins bg color

taranchauhan:
**Checklist**:
* [x] Documentation
* [x] Tests
* [x] Ready to be merged

### Update HiddenText to align with [Details component](https://design-system.service.gov.uk/components/details/) margins and focus background colour

- Add hoc as a package to HiddenText
- Add whiteSpace on details only and remove from inset text to align alpha/govuk-frontend
- Spread props to details instead of InsetText so that the withWhiteSpace can be applied.
- Change background colour to FOCUS_COLOUR based on new GDS design system at https://design-system.service.gov.uk/components/details/
- Update HiddenText snapshot

### DIFF
--- a/components/hidden-text/package.json
+++ b/components/hidden-text/package.json
@@ -3,8 +3,9 @@
   "version": "0.2.5",
   "dependencies": {
     "@govuk-react/constants": "^0.2.2",
-    "@govuk-react/inset-text": "^0.2.5",
-    "@govuk-react/paragraph": "^0.2.5",
+    "@govuk-react/hoc": "^0.2.4",
+    "@govuk-react/inset-text": "^0.2.4",
+    "@govuk-react/paragraph": "^0.2.4",
     "govuk-colours": "^1.0.3"
   },
   "peerDependencies": {
--- a/components/hidden-text/src/index.js
+++ b/components/hidden-text/src/index.js
@@ -46,6 +46,7 @@ const StyledSummary = styled('summary')({
  *
  * ### References
  * - https://govuk-elements.herokuapp.com/typography/#typography-hidden-text
+ * - https://github.com/alphagov/govuk-frontend/blob/master/src/components/details/_details.scss
  */
 const HiddenText = ({ summaryText, children, ...props }) => (
   <details {...props}>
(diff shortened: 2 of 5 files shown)


components/hidden-text/package.json
```diff
@@ -3,8 +3,9 @@
   "version": "0.2.5",
   "dependencies": {
     "@govuk-react/constants": "^0.2.2",
-    "@govuk-react/inset-text": "^0.2.5",
-    "@govuk-react/paragraph": "^0.2.5",
+    "@govuk-react/hoc": "^0.2.4",
```
Loque-: probably need to rebase the branch off master as it seems to regress the version numbers here?

components/hidden-text/src/index.js
```diff
@@ -23,8 +24,9 @@ const StyledSummary = styled('summary')({
     color: LINK_HOVER_COLOUR,
   },
   ':focus': {
-    outline: `${FOCUS_WIDTH} solid ${FOCUS_COLOUR}`,
-    outlineOffset: '-1px',
+    outline: `calc(${FOCUS_WIDTH} + 1px) solid ${FOCUS_COLOUR}`,
```
Loque-: Shame to have to use calc here for this, I thought maybe the calculation could be done with JS and the result used here, but you will have to strip the `px` off the values and then re-add them... bit of a schlep?

components/hidden-text/src/index.js
```diff
@@ -45,19 +47,21 @@ const StyledSummary = styled('summary')({
  * ### References
  * - https://govuk-elements.herokuapp.com/typography/#typography-hidden-text
  */
-const HiddenText = ({ summaryText, ...props }) => (
-  <details>
+const HiddenText = ({ summaryText, children, ...props }) => (
+  <details {...props}>
```
Loque-: nice, great adjustment

components/hidden-text/src/index.js
```diff
@@ -45,19 +47,21 @@ const StyledSummary = styled('summary')({
...
     <StyledSummary><StyledSpan>{summaryText}</StyledSpan></StyledSummary>
-    <InsetText isNarrow {...props} />
+    <InsetText mb={0} isNarrow>{ children }</InsetText>
   </details>
 );
 
 HiddenText.defaultProps = {
+  children: undefined,
```
Loque-: Would be great to add docs to these props and then generate the documentation

components/hidden-text/package.json
Loque-: Really odd - wonder if this is project wide ~ I might have mashed something in the release. Thanks for the update.

Loque-:
odd, travis slipped up on the snapshots, I might try regenerating them here.

PR #353, opened 2018-06-28
Feature/update documentation

### LINKED ISSUE #338: Use of <strong> for bold text is not accessibility-friendly
I have noticed that most (possibly all) cases where text is supposed to be **bold**, this is achieved through a `<strong>` tag. This includes the markdown implementation and some components where the text is always bold.

Screen readers interpret `<strong>` tags as not just bold, but extremely emphasised, so the volume increases dramatically when the reader encounters these tags.

This detracts from the experience and user-friendliness for accessibility users, so I have been told that this is considered a bad practice.

Bold styling should be applied instead with CSS alone.

Loque-:
Update documentation for using Layout components. Minor improvements to Layout components. Removes standalone `Layout` component from storybook as it is no longer required and added inline notes about why. Regenerates documentation.

Two sneaky non-PR-related adjustments;
- Added note about using bold text (fixes #338)
- renamed `...rest` to `...props` to follow project standard

* [x] Documentation
* [x] Tests
* [x] Ready to be merged


Loque-:
Thanks Gavin, I'm just going to nudge the build via a commit to ensure Travis passes oki - think it must have hit a network error.

Loque-:
Yes it would be we don’t want to add dependacies required to render stories in the components package.json (as they are not important to the component) - there is a rule for this in our .eslintrc file but for some reason it still complains if I import from ‘govuk-react’.

Maybe it is should live in the devDependancies - but arguably, not really required for developing the component itself as it is limited to the story file only... I think I’d be happy placing it there however.

PR #361, opened 2018-07-03
add tabbed-nav component

hom-bahrani:
@Loque- first time I'm using storybook so may not have got that bit right, would be grateful for your guidance when you get a chance to review this. thanks :) 

**Checklist**:

* [ x] Documentation
* [ x] Tests
* [ x] Ready to be merged


components/tabbed-nav/src/index.js
```diff
@@ -0,0 +1,104 @@
...
+const NavUl = styled('ul')`
+    list-style: none;
+    margin: 0;
+    border-top: 0;
+    padding: 0;
+    border-bottom: 2px solid #bfc1c3; 
+`;
+
+const Navli = styled('li')`
+  float: left;
```
Loque-: Lets use display: flex for this instead! You can remove all the floats & clears and add this to the `UL`: 
```
display: flex;
align-items: flex-end;
```

Should do it?

components/tabbed-nav/src/index.js
```diff
@@ -0,0 +1,104 @@
+import React from 'react';
+import styled from 'react-emotion';
+import PropTypes from 'prop-types';
+
+// styled components
+const NavNavContainerli = styled('div')`
```
Loque-: I don't think you need this element? By default `div` has no `margin`, `padding` and has `display: block;` - I think this component would work if you removed this container completely toooooo

components/tabbed-nav/src/index.js
```diff
@@ -0,0 +1,104 @@
...
+    text-transform: none;
+    cursor: pointer;
+    display: block;
+    border-bottom: none;
+    float: left;
+    height: 3.25em;
+    line-height: 3.25em;
+    margin-bottom: -2px;
+    padding: 0 40px;
+    border-top: ${props => props.active ? '2px solid #bfc1c3' : 'none'};
```
Loque-: There are more elegant ways of doing this - if you reference some of the other components like `Button` you'll see it de-structures props.

Also, we will need to swap these colour vars for govuk-colours, which is reference somewhere (check `Button` again for that too).

Loque-:
Nice one - I'll check it out and verify storybook in a bit. Left some initial comments to help, let me know if I can help any further on slack toooooo - I'm currently a bit confused over how it has different content for each tab but that might clear itself when I try it out.

Loque-:
Just to share discussion away from this component incase someone else picks it up, Homam and I discussed adjusting it to work a bit more like;

```
const childComponentOne = () => (
  <div>
    Second child component
  </div>
);

const tabbedStuff = [
  { title: 'some tab title', body: childComponentOne }
  { title: 'some tab title', body: 'some body content, need to check content is supported like this, maybe markdown?' }
  { title: 'some tab title', body: 'some body content, need to check content is supported like this, maybe markdown?' }
]

<TabbedNav tabs={tabbedStuff} />
```

Closing for the mo!

PR #368, opened 2018-07-13
Header from H2 -> H3 in error summary page

anishamalde:
**Checklist**:

* [ x ] Documentation
* [ x ] Tests
* [ x ] Ready to be merged 

screen shot 2018-07-13 at 10 24 54

### DIFF
--- a/components/error-summary/src/index.js
+++ b/components/error-summary/src/index.js
@@ -117,7 +117,7 @@ const ErrorSummary = ({
 }) => (
   <StyledErrorSummary tabIndex={-1} {...props}>
     <Header level={2}>{ heading }</Header>
-    { description && <Paragraph mb={2}>{ description }</Paragraph> }
+    { description && <Paragraph mb={3}>{ description }</Paragraph> }
     { errors.length > 0 &&
       <UnorderedList listStyleType="none">
         { errors.map(error => (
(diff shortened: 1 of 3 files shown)


packages/hoc/src/withWhiteSpace/__snapshots__/test.js.snap
```diff
@@ -20,6 +20,7 @@ exports[`withWhiteSpace matches wrapper snapshot 1`] = `
 
 <InputField
   className="emotion-0"
+  hint={null}
```
Loque-: Not sure how this got into the snapshots but looks like a mistake (as the code change does not effect this file) - if you could revert this change that would be awessssoommmeee.

components/error-summary/src/index.js
```diff
@@ -115,29 +115,29 @@ const StyledErrorSummary = styled('div')({
...
-          <ListItem key={error.targetName}>
-            <StyledErrorText
-              tabIndex={-1}
-              onClick={() => onHandleErrorClick(error.targetName)}
-            >
-              {error.text}
-            </StyledErrorText>
-          </ListItem>
+    <StyledErrorSummary tabIndex={-1} {...props}>
+      <Header level={3}>{heading}</Header>
```
Loque-: This is the only code change, but something has adjusted all the formatting - if you could revert the formatting changes, so the code change (from `2` - `3`) as the only change that would be mega.

Loque-:
nice one!

PR #374, opened 2018-07-20
Feature/storybook enhancements

gavinorland:
Right, here's my solution for https://github.com/UKHomeOffice/govuk-react/issues/317 to:

- Get story categories to be listed alphabetically
- Get a custom _Welcome_ story at the top
- Get this to render the `readme` file!

See what you guys think and please feel free to suggest - or commit - if you can see any better way of doing this :)

### DIFF
--- a/packages/storybook/.storybook/addons.js
+++ b/packages/storybook/.storybook/addons.js
@@ -1,3 +1,4 @@
 import '@storybook/addon-actions/register';
 import '@storybook/addon-links/register';
 import '@storybook/addon-knobs/register'
+import '@storybook/addon-options/register';
--- a/packages/storybook/.storybook/addons.js
+++ b/packages/storybook/.storybook/addons.js
@@ -1,4 +1,4 @@
 import '@storybook/addon-actions/register';
 import '@storybook/addon-links/register';
-import '@storybook/addon-knobs/register'
+import '@storybook/addon-knobs/register';
 import '@storybook/addon-options/register';
--- a/packages/storybook/.storybook/config.js
+++ b/packages/storybook/.storybook/config.js
@@ -1,5 +1,11 @@
 import { configure } from '@storybook/react';
 import 'react-chromatic/storybook-addon';
+import { setOptions } from '@storybook/addon-options';
+
+setOptions({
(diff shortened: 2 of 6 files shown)


packages/storybook/.storybook/addons.js
```diff
@@ -1,3 +1,4 @@
 import '@storybook/addon-actions/register';
 import '@storybook/addon-links/register';
 import '@storybook/addon-knobs/register'
+import '@storybook/addon-options/register';
```
stevesims: i think our rules say that there should be a newline here

packages/storybook/.storybook/config.js
```diff
@@ -1,5 +1,11 @@
 import { configure } from '@storybook/react';
 import 'react-chromatic/storybook-addon';
+import { setOptions } from '@storybook/addon-options';
+
+setOptions({
+  name: 'govuk-react',
+  sortStoriesByKind: true,
```
stevesims: noice :+1:

packages/storybook/src/stories/index.js
```diff
@@ -1,5 +1,11 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
 import 'normalize.css';
 import './styles.css';
+import readme from '../../../../README.md';
+
+// eslint-disable-next-line react/no-danger
+storiesOf(' Welcome', module).add('to govuk-react', () => <div style={{ padding: '10px' }} className="markdown-body" dangerouslySetInnerHTML={{ __html: readme }} />);
```
stevesims: is this not just setting raw markdown as the HTML content?

in other words, it feels like this won't actually be rendering the markdown content

we can probably use `react-markdown` to ensure that the MD gets rendered as proper HTML
that would remove the need for the `react/no-danger` lint disable

gavinorland: Yes - I had a look at using `react-markdown` but the funny thing is when the file is displayed it is _not_ markdown but is already rendered as HTML! However, the HTML is just rendered into the page unless we use `dangerouslySetInnerHTML`.

Fire it up and take a look...

PR #379, opened 2018-07-27
Feature/improve grid col

### LINKED ISSUE #377: More flexibility with GridCol
Currently `GridCol` supports flags for 1/4, 1/3, 1/2 and 2/3 width columns.

By default a column is 100% wide if a width flag is not provided.  Therefore, as there is not a flag to support it, it is not possible to make a 3/4 width column.  It is therefore suggested that we should add support for a `columnThreeQuarters` flag.

Additionally there is no "auto" width column to make use of the remaining flex area in a `GridRow`.  To provide this we could change the default column width be `auto` with `flex: 1` set, and have the explicit widths set `flex: 0`.  (This change would arguably negate the need for an explicit `columnThreeQuarters`.)  Alternatively we could provide an explicit `columnAuto` flag to set `width: auto; flex: 1;`.

It is also not possible to do columns of 1/6th or 5/6th (both of which would be useful in our application). 
 Adding support for `columnOneSixth` and `columnFiveSixths` would also be cool, but as those are a less common requirement and may not form part of the main GDS styling options we could instead add support for those in our other component library.

Loque-:
Fixes #377 

- Removes 100% width
- Sets flex grow value to 0 if element has a requested col width, otherwise it is set to 1, allowing it to auto-expand
- Reduces a lot of un-required CSS by only rendering css props if specific props are passed
- Only applies desired margin values to avoid overriding values potentially requested elsewhere (like margin top or bottom

* [x] Documentation
* [x] Tests
* [x] Ready to be merged

### DIFF
--- a/components/grid-col/src/index.js
+++ b/components/grid-col/src/index.js
@@ -25,13 +25,14 @@ const StyledColumn = styled('div')(
       },
     },
   },
-  ({ hideContent }) => ({
-    textIndent: hideContent ? '-999em' : undefined,
-    backgroundColor: hideContent ? '#7DADD3' : undefined,
-    backgroundImage: hideContent
-      ? 'repeating-linear-gradient(180deg, #7DADD3, #7DADD3 15px, #B7CFE1 15px, #B7CFE1 30px)'
-      : undefined,
-  }),
+  ({ hideContent }) => {
+    if (!hideContent) { return false; }
+    return ({
+      textIndent: '-999em',
+      backgroundColor: '#7DADD3',
+      backgroundImage: 'repeating-linear-gradient(180deg, #7DADD3, #7DADD3 15px, #B7CFE1 15px, #B7CFE1 30px)',
+    });
+  },
   (...args) => {
     let widthValue = 'auto';
     let hasRequestedWidth = false;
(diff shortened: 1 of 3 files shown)


components/grid-col/src/index.js
```diff
@@ -20,33 +25,31 @@ const StyledColumn = styled('div')(
...
-  }),
+  ({ hideContent }) => {
+    if (!hideContent) { return false; }
+    return ({
+      textIndent: '-999em',
+      backgroundColor: '#7DADD3',
+      backgroundImage: 'repeating-linear-gradient(180deg, #7DADD3, #7DADD3 15px, #B7CFE1 15px, #B7CFE1 30px)',
+    });
+  },
+  (...args) => {
```
stevesims: Rather than `...args` here this could have just been `props` and instead of `args[0]` below that could have been `prope` too

Would be a little easier to understand 

You’re not looking at any other argument being passed in so the spread isn’t needed

components/grid-col/src/stories.js
```diff
@@ -110,3 +110,17 @@ examples.add('Quarters', () => (
     </GridRow>
   </Fragment>
 ));
+
+examples.add('One Quarter and autoFill', () => (
+  <Fragment>
```
stevesims: This Fragment isn’t needed as it contains only a single GridRow

Loque-: Good point, it's not needed for any of the others too, removing.

PR #382, opened 2018-07-30
Apply more sensible minHeight value of '10px' as a fallback

### LINKED ISSUE #381: Apply more sensible minHeight for LoadingBox to prevent unnecessary white space
1. min-height of '100px' was added to LoadingBox so that for very short components the Loading spinner would still be of a sufficient height to be visible. 
2. However, this breaks inside the web app because there is extra whitespace due to the min-height of '100px' being applied.

**Potential Solution**: Apply more sensible min-height of 10px as a fall back so at least the spinner will be displayed.

taranchauhan:
**Checklist**:
* [x] Documentation
* [x] Tests
* [x] Ready to be merged

Solution: Apply more sensible min-height of 10px as a fall back so at least the spinner will be displayed.

closes #381

### DIFF
--- a/components/loading-box/src/index.js
+++ b/components/loading-box/src/index.js
@@ -8,13 +8,11 @@ import { BLACK, WHITE } from 'govuk-colours';
 
 const spinnerClassName = 'icon-loading';
 
-const StyledContainer = styled('div')(({
-  loading,
-}) => ({
+const StyledContainer = styled('div')({
   position: 'relative',
   paddingBottom: '2px',
-  minHeight: loading ? '100px' : undefined,
-}));
+  minHeight: '10px',
+});
 
 const Innerwrap = styled('div')(({
   timeIn,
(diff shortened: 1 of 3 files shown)


components/loading-box/src/index.js
```diff
@@ -8,11 +8,13 @@ import { BLACK, WHITE } from 'govuk-colours';
...
 
-const StyledContainer = styled('div')({
+const StyledContainer = styled('div')(({
+  loading,
+}) => ({
   position: 'relative',
   paddingBottom: '2px',
-  minHeight: '100px',
-});
+  minHeight: loading ? '100px' : undefined,
```
stevesims: cool - i did think of this... but i don't think it's the right solution here

we have use-cases where the loading box is wrapped around a drop-down menu where this will force the container to be too tall for the duration of when the data is loading

IMHO we're better off either removing this min-height, and relying on the component that's using the loading box to specify a min-height on the contents, or setting it to a much smaller value like 10px as a fall-back "make sure we get to display *something*"

Issue #383, opened 2018-08-07
Update the package.jsons of all components


gavinorland:
The `description` field should outline what the particular component is/does.

PR #388, opened 2018-08-22
Add homepages and descriptions to component package.jsons

gavinorland:
Hitherto, each package inside govuk-react had a generic homepage and description. This PR addresses #383 by adding:

- a component specific URL
- a basic description of what the component is/does

...inside each `package.json`.

I haven't edited the authors as with govuk-frederic (usually "Alasdair McLeay") - this could be done later.

**Checklist**:

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged


Loque-:
Shouldn't the homepage be the root folder for the specific component on github and not the npm path? https://docs.npmjs.com/files/package.json#homepage

e.g; https://github.com/UKHomeOffice/govuk-react/tree/master/components/back-link

Wondering if it should be sympathetic to its version number, but probably overkill, unless an easy implementation is available.

gavinorland:
Ah yes, you're right. I'll tweak that on each - and the same for Frederic.

gavinorland:
As with Frederic, checks don't seem to be completing here but perhaps an admin can approve and force through the merge anyway unless anyone would like to suggest any edits.

gavinorland:
Regex has made quick work of removing all the anchors from the URLs so this is ready for another look!

gavinorland:
Thanks but it's saying checks haven't completed again so I can't merge.

Issue #339, opened 2018-06-20
Triangle on HiddenText is being removed in some browsers


Joe-Brady:
Looks like there is a feature of the `<summary>` element that allows you to remove the twisty triangle by changing the `display` type in CSS.

However, this feature isn't yet in Chrome by the looks of it, so we only just noticed that the triangle is missing in Firefox, for example!

Looks like the `display: inline-block` is the culprit. Its default is `list-item`:

"Per the HTML specification, the default style for `<summary>` elements includes `display: list-item`. This makes it possible to change or remove the icon displayed as the disclosure widget next to the label from the default, which is typically a triangle."

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary

<img width="730" alt="screen shot 2018-06-20 at 15 06 16" src="https://user-images.githubusercontent.com/20062773/41664016-a999e39c-749c-11e8-9190-c5b857bbee95.png">
<img width="1573" alt="screen shot 2018-06-20 at 15 06 39" src="https://user-images.githubusercontent.com/20062773/41664017-a9c9e290-749c-11e8-9b0b-db652623732f.png">

igloosi:
just to note that summary implementation in GOV.UK Frontend is not affected by this change

PR #389, opened 2018-08-22
Proposed fix for #339

gavinorland:
Having replicated #339 in Firefox (Mac) locally, this is a proposed fix - simply removing the display statement - but of course this will only be acceptable if it doesn't have any adverse effects where the component is used.

**Checklist**:

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged


gavinorland:
Not quite sure how we can the required tasks on this to complete - or perhaps Steve will be able to force through a merge anyway.

Issue #358, opened 2018-06-29
checkbox is selected when clicking outside label/input


hom-bahrani:
odd behaviour with the checkbox, observed by our QA. When you click to the right of the checkbox much further to the right of the page the checkbox can be selected / deselected. this results in accidental selection
![checbox](https://user-images.githubusercontent.com/8465628/42134795-09915a50-7d3a-11e8-89e9-642b89d84697.gif)

Loque-:
Ah nice spot, classic one this - need to use `display: inline-block;` here: https://github.com/UKHomeOffice/govuk-react/blob/master/components/radio/src/index.js#L13

We can add a quick fix for this, but, we might also be able to fly through all the form elements next week also. If anyone needs something immediately this would work;

```js
import styled from 'react-emotion';
import Radio from '@govuk-react/radio';

const RadioField = styled(Radio)({
  display: 'inline-block',
})
```

PR #392, opened 2018-08-24
Fix checkbox click area (#358)

gavinorland:
Fix for #358: If we use `block` here then labels span the width of the page. If we use `inline-block` then they collapse, but wrap together side by side. If we use `table` then they collapse but still stack under one another = win!

**Checklist**:

* [ ] Documentation N/A
* [x] Tests
* [x] Ready to be merged

### DIFF
--- a/components/checkbox/src/index.js
+++ b/components/checkbox/src/index.js
@@ -10,7 +10,8 @@ import {
   NTA_LIGHT,
 } from '@govuk-react/constants';
 
-const StyledCheckbox = styled('div')({
+const StyledCheckbox = styled('label')({
+  display: 'table',
   position: 'relative',
   padding: '0 0 0 38px',
 });
@@ -26,10 +27,10 @@ const StyledInput = styled('input')(
     margin: 0,
     zoom: 1,
     opacity: 0,
-    ':checked + label:after': {
+    ':checked + span:after': {
       opacity: 1,
     },
-    ':focus + label:before': {
+    ':focus + span:before': {
       boxShadow: `0 0 0 4px ${YELLOW}`,
     },
   },
@@ -42,7 +43,7 @@ const StyledInput = styled('input')(
   }),
 );
 
-const StyledLabel = styled('label')({
+const StyledLabel = styled('span')({
   fontFamily: NTA_LIGHT,
   fontWeight: 400,
   textTransform: 'none',
@@ -54,7 +55,7 @@ const StyledLabel = styled('label')({
   },
   cursor: 'pointer',
   padding: '8px 10px 9px 12px',
-  display: 'inline-block',
+  display: 'block',
   color: `${BLACK}`,
   '::before': {
     content: "''",
@@ -112,12 +113,13 @@ const StyledLabel = styled('label')({
  *
  */
 const Checkbox = ({
-  children, className, id, ...props
+  children, className, ...props
 }) => (
   <StyledCheckbox className={className}>
-    <StyledInput id={id} type="checkbox" {...props} />
-    <StyledLabel htmlFor={id}>{children}</StyledLabel>
-  </StyledCheckbox>);
+    <StyledInput type="checkbox" {...props} />
+    <StyledLabel>{children}</StyledLabel>
+  </StyledCheckbox>
+);
 
 Checkbox.defaultProps = {
   className: undefined,
@@ -132,10 +134,6 @@ Checkbox.propTypes = {
    * CSS Classname for outermost container
    */
   className: PropTypes.string,
-  /**
-   * ID - for label to use
-   */
-  id: PropTypes.string.isRequired,
 };
 
 export default withWhiteSpace({ marginBottom: 2 })(Checkbox);
--- a/components/checkbox/src/index.js
+++ b/components/checkbox/src/index.js
@@ -11,7 +11,7 @@ import {
 } from '@govuk-react/constants';
 
 const StyledCheckbox = styled('label')({
-  display: 'table',
+  display: 'inline-block',
   position: 'relative',
   padding: '0 0 0 38px',
 });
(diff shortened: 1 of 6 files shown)


components/checkbox/src/index.js
```diff
@@ -11,7 +11,7 @@ import {
 } from '@govuk-react/constants';
 
 const StyledCheckbox = styled('label')({
-  display: 'block',
+  display: 'table',
```
Loque-: I'm not so sure using display: table is a good idea, is there a fix that feels like less of a hack?

Loque-: I only say this as it is rather unconventional and could have unwanted side effects/trip up future development or people implementing it?

penx: also the idea is to mirror GDS styles where possible unless there is a good reason not to, preferably related to React/CSSinJS

I'd recommend wherever we are not mirroring the styles there should be a comment inline explaining the choice

Loque-: That's a good shout actually - using flexbox to solve some problems might be a nice improvement and is easy to add a comment to explain why also.

gavinorland: Yes, it is a bit unconventional.. just works! :stuck_out_tongue:

If we go with Flexbox then I think all we need to do is add `display: flex;` to whatever contains the checkboxes. Is there such a container/could there be?

Loque-: My note about flexbox was generic and not specific to this (as there are previous instances where flexbox had been used and was not on GDS) - I might be able to have a look on Thursday! If I can before I will mention on here!

penx:
When adding bugfixes can you try to think if there’s a unit test that could show the issue beyond a snapshot test?

penx:
Also I’m not sure this is something we need to fix. 

What is the default html behaviour?

What is the govuk-frontend and predecessor behaviour?

Loque-:
It's caused by the wrapping label element, which I believe can be fixed by setting inline-block on it, not sure why that didn't fix it if that was attempted?

Loque-:
It might be worth considering using flexbox here as it allows better customisation of the contained elements?

gavinorland:
If we set `inline-block` just on the `label` element then it stops the full-width block behaviour but does mean all the labels wrap together on the same line.

Loque-:
I can't see why you would need a `LabelContainer` using flexbox?

Loque-:
display flex and flex direction column on the label would work?

Loque-:
Yea it doesn't need an additional container, but it suffers from the same problem that display block does.

If we are really going to copy GDS CSS and we want things like this to work, we will probably need to copy the markup as well (no biggy really imho), as it is simply the behaviour of the label element that is kicking off.

gavinorland:
Yes, having now seen the govuk-frontend examples I notice each component has a wrapping div which contains the input and label, whereas those in govuk-react don't include this. It also uses a label, not a span, for the text, and sets the `for` attribute on it and an `id` on the checkbox to link the two.

I've just added a commit to make the structure of ours resemble this. The tests could probably be improved a bit but take a look, also in Storybook - working well! n.b. It does require that wherever this is used an `id` is sent in.

penx:
Just remembered labels are actually one place where we're intentionally doing something to govuk-frontend.

In govuk-frontend, labels don't wrap form elements, but this requires form elements to have unique IDs.

As React doesn't require form elements to have IDs, and due to the complexity of producing consistent IDs for form fields that would (a) work with universal rendering (b) be easy for the parent application to implement in all scenarios, we decided to make labels wrap the form element.

Perhaps that's what the issue is here?

gavinorland:
Yes, that's certainly related, @penx, as per my commit above. (btw I just forced pushed an update to pass tests - ESLint does not pick up all files with the current glob.)

If we want to avoid `id`s and keep the label on the outside, containing the checkbox, we can give this a bit more thought - probably just add a wrapping div.

penx:
> If we want to avoid ids and keep the label on the outside

Yep, well not avoiding them altogether, just not making them mandatory in order for label association to work.

gavinorland:
Sometimes the simplest things are best! Check out the latest commit compared with the previous fix. This works no problem so perhaps we are there.

Loque-:
I'm not a fan of anonymous divs personally, but not enough to fight for it - I'd personally find a way to link the label and the input in a decent, instance related way, but yea, each to their own! : )

penx:
> I'm not a fan of anonymous divs personally

Are you against what we are trying to achieve, i.e. an inline block inside a block? If not, an anonymous div seems like the best approach, so not sure why you'd be against it for this usage.

Yes I think the className should be on the outer div.

Happy to recommend we pause this until we can all discuss in person if you're not happy :-)

gavinorland:
Hi. Perhaps the above will be best - we end up with this:

```
<StyledContainer className={className}>
  <StyledLabel>
    <StyledInput type="checkbox" {...props} />
    <StyledSpan>{children}</StyledSpan>
  </StyledLabel>
</StyledContainer>
```

Loque-:
I'm only against it if it doesn't have the `className` prop on it as it is no longer anonymous!

stevesims:
I've taken a look at this too now...

I'm not a fan of the approach taken.  Adding in more markup to work around a problem always feels slightly dodgy to me.

Additionally I disliked the original styling solution too - as overriding font choices wasn't possible.

This is my preferred approach: https://codesandbox.io/s/43p4wm4p89

gavinorland:
This seems to use `label`'s default `inline-block` behaviour and `<br>` tags to achieve line breaks between each label? However when I try two of them the display isn't looking right. https://codesandbox.io/s/8lpoy2z5yl

Maybe this is not an issue if these labels are always expected be used inside `div`s.

gavinorland:
Steve's confirmed that with his suggested implementation the labels would align on one line due to the `inline-block` behaviour of labels, and line breaks would be achieved by using container divs at implementation level.

This doesn't match the govuk-frontend structure, which I gather is something @penx wanted to do (that has a wrapping div) - but then that is also different with its `for` attribute approach. :)

My last suggestion has been approved but required tests haven't completed so I can't merge. @stevesims Maybe you could implement your suggestion here instead or merge this.

penx:
@gavinorland closing this for now. Feel free to revisit comments above and reopen later

PR #397, opened 2018-09-04
Bring MultiChoice test coverage to 100%

gavinorland:
Also re-wrote tests to be more explicit and to no longer send in unexpected `error` prop.

When this and the [Paragraph test](https://github.com/UKHomeOffice/govuk-react/pull/396) are merged it'll bring govuk-react coverage to 100%.

**Checklist**:

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged

### DIFF
--- a/components/multi-choice/src/stories.js
+++ b/components/multi-choice/src/stories.js
@@ -1,22 +1,19 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
+import { withKnobs } from '@storybook/addon-knobs/react';
 import { WithDocsCustom } from '@govuk-react/storybook-components';
 
-import Radio from '@govuk-react/radio';
-import MultiChoice from '.';
+import MultiChoiceWithKnobs,
+{ MultiChoiceWithKnobsHint, MultiChoiceWithKnobsError } from './fixtures';
+
 import ReadMe from '../README.md';
 
 const stories = storiesOf('Form/MultiChoice', module);
-
+stories.addDecorator(withKnobs);
 stories.addDecorator(WithDocsCustom(ReadMe));
+stories.add('Component default', () => <MultiChoiceWithKnobs />);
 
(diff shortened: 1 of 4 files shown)


components/multi-choice/src/test.js
```diff
@@ -1,38 +1,33 @@
...
-  const meta = {
-    touched: true,
-    error: example,
-  };
-  const wrapper = (
-    <MultiChoice meta={meta} hint={example} label={example}>
-      {example}
-    </MultiChoice>);
+  it('renders with a label', () => {
+    const wrapper = mount(<MultiChoice label="Label">Example</MultiChoice>);
```
penx: Ideally the different fixtures used by tests should be in `fixtures.js` and used by both `stories.js` and `test.js` to ensure that what we have covered by snapshot tests also appears in our visual regression tests.

penx: would these new snapshot tests be useful to have on storybook examples/chromatic?
If so, please add them via fixtures.js
If not, why not?

gavinorland: 👍 I'll take a look at this and see if I can break out anything shared.

components/multi-choice/src/stories.js
```diff
@@ -1,22 +1,19 @@
...
-      Yes
-    </Radio>
-    <Radio name="group1" inline>
-      No
-    </Radio>
-  </MultiChoice>
-));
+const examples = storiesOf('Form/MultiChoice/Examples', module);
+examples.addDecorator(withKnobs);
+examples.add('With a hint', () => <MultiChoiceWithKnobsHint />);
```
penx: does`examples.add('With a hint', MultiChoiceWithKnobsHint);` work?

penx: approved other than this

PR #401, opened 2018-09-05
Exempt component docs for Chromatic userAgent

gavinorland:
**Checklist**:

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -7,6 +7,7 @@ module.exports = {
     "flowtype"
   ],
   "env": {
+    "browser": true,
     "es6": true
   },
   "rules": {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -7,7 +7,6 @@ module.exports = {
     "flowtype"
   ],
   "env": {
-    "browser": true,
     "es6": true
   },
   "rules": {
--- a/packages/storybook-components/src/WithDocsCustom.js
+++ b/packages/storybook-components/src/WithDocsCustom.js
@@ -16,12 +16,12 @@ const PreviewComponent = ({ children }) => (
   </div>
 );
 
-const WithDocsCustom = withDocs({
-  PreviewComponent,
-});
+const withDocsCustom = !navigator.userAgent.match(/Chromatic/) ?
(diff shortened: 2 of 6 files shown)


.eslintrc.js
```diff
@@ -7,6 +7,7 @@ module.exports = {
     "flowtype"
   ],
   "env": {
+    "browser": true,
```
penx: you are changing the eslint settings for the whole project, which should be runnable on the server (universal React) to allow browser variables

packages/storybook-components/src/WithDocsCustom.js
```diff
@@ -16,12 +17,10 @@ const PreviewComponent = ({ children }) => (
...
-  PreviewComponent,
-});
+const withDocsCustom = (readme, ...rest) => withDocs({ PreviewComponent })(navigator.userAgent.match(/Chromatic/) ? '' : readme, ...rest);
 
 PreviewComponent.propTypes = {
   children: PropTypes.node.isRequired,
 };
 
-export default WithDocsCustom;
+export default withDocsCustom;
```
penx: Please could you or @Loque- create a ticket to address HOCs being named appropriately and not filed under 'Components'
- [ ] ticket created

.eslintrc.js
gavinorland: Hmm.. I'm not sure how we can have environment-dependent ESLint rules, unless there is some kind of environment variable we can check. We _can_ enable the browser environment inline using:

`/* eslint-env browser */`

But I think our best bet here is probably to just disable ESLint on the relevant lines using:

`// eslint-disable-next-line no-undef`

Let me know your thoughts.

penx: There are a few ways to add eslint rules that only apply to certain file names or folders

gavinorland:
We have approved all of the Chromatic screenshots (there's a batch approve button!). I don't think they their visuals will changes even if we make another commit e.g. related the ESLint stuff above - but let's see. I think we can easily batch re-approve again if necessary.

PR #403, opened 2018-09-07
Remove react-router dependency

### LINKED ISSUE #282: We should be router agnostic
Right now we have a dependency on `react-router-dom` - we should remove this.

Including it means we're not agnostic about routing and whilst our library users may also be using `react-router` they may be on a different version.  Additionally it stops us from being react-native compatible.

It's suggested that in the few components that require the `Link` component from `react-router-dom` (as I believe this is the only component that's being used) that we support passing in the `Link` component and fall-back to using plain `a` tags

penx:
Parent application needs to specify a custom link renderer if they want to use a react router Link. This would be a breaking change for anyone using the paragraph component and expecting it to work with React Router out of the box

Fixes #282 

related #223

**Checklist**:

* [x] Documentation
* [x] Tests N/A
* [ ] Ready to be merged


penx:
Next publish after merging this should update to v0.3.0

lerna.json
```diff
@@ -2,7 +2,7 @@
   "lerna": "2.9.0",
   "npmClient": "yarn",
   "useWorkspaces": true,
-  "version": "0.2.11",
+  "version": "0.3.0-prerelease",
```
Loque-: Is this intentional?

penx: yes, because it ensures that the next npm release would be 0.3.0.

As this PR contains breaking changes, we want to ensure the next version after merge is 0.3.0.

penx: you can test this in a new folder:

```
mkdir test-prerelease
cd test-prerelease
npm init 0.3.0-prerelease
cat package.json
npm version patch
cat package.json
```

penx:

Built with commit b458f92a017b42022a3c1d27f8103fccc97576b6

https://app.netlify.com/sites/affectionate-nobel-f3a8ec/deploys/5bd2f1c73813f02551da37a7

PR #406, opened 2018-09-11
Fix header stacking

anishamalde:
**Checklist**:

* [x ] Documentation
* [x ] Tests
* [x] Ready to be merged 

Fixed version- With flex-start
screen shot 2018-09-10 at 14 54 36

Current with flex-end
screen shot 2018-09-11 at 10 39 48


Loque-:
Thank you, just checking this out - in the future, if you make PR's from this repo instead of a fork I can check them out locally, test, and collaborate easily toooooo

Loque-:
@anishamalde looks like it is failing due to codecov: https://travis-ci.org/UKHomeOffice/govuk-react/builds/427131575#L1421 - might need to wait for @penx to check tomorrow unless it starts working after some further adjustments : )

I'm using this https://design-system.service.gov.uk/components/header/with-service-name-and-navigation/index.html as a reference for service title and subnavigation (from this page: https://design-system.service.gov.uk/components/header/). I might try to ensure the mobile menu is rendered on the left (as per the spec), whilst also rendering the service title on the left (as per this change)... just looking at the component now.

penx:
It's failing because the branch is on a different GitHub account. This means travis doesn't supply the env variables to the build script and then codecov fails.

We either need to ensure all PRs come from the UKHomeOffice org, or else skip this step if the env variable is not available.

Loque-:
Superseded by https://github.com/UKHomeOffice/govuk-react/pull/411, cheers!

PR #408, opened 2018-09-13
Feature/default value to date

anishamalde:
**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged

### DIFF
--- a/components/date-input/src/__snapshots__/test.js.snap
+++ b/components/date-input/src/__snapshots__/test.js.snap
@@ -177,9 +177,9 @@ exports[`DateInput matches wrappersnapshot: wrapper mount 1`] = `
     className="emotion-29"
     defaultValues={
       Object {
+        "defaultDay": undefined,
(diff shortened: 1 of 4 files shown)


penx:
I was a bit concerned that defaultValue is needed as this is only supposed to be for uncontrolled components AFAIK, and generally developers shouldn't be using uncontrolled components - https://reactjs.org/docs/uncontrolled-components.html

I then took a look at the file and don't see value being passed to the inputs, so I don't see how DateInput can be used as a controlled component in the first place.

So from what I can tell, there is a much bigger issue here, in that DateInput cannot be used as a controlled input. If this was fixed, then the parent component could be controlling the default.

How is this component being used at present? It seems incomplete.

penx:
As it stands I have no issue with the proposed change if this component is intended to be used uncontrolled, I am just worried this component is even being used at the moment :-)

PR #443, opened 2018-10-29
Make grid consistent with govuk-frontend

### LINKED ISSUE #432: GridCol should have flex-shrink 0 set
We have a use of `GridCol` on a 2 column page where the LH column uses the `columnOneQuarter` flag, and the RH column expands to fill the space.  In the LH column is placed a table of contents, styled using `position: fixed` so that it remains in place whilst the page is scrolled.

In this scenario the RH column will grow to almost the full width of the page, and will overlap the table of contents in the LH column.

This can be fixed by applying a `flex-shrink: 0` to elements that have a fixed width (i.e. those that use one of the width setting flags).

NB applying `flex-shrink: 0` cannot be a blanket thing on columns as setting it on a "use all remaining space" column such as the RH column in this example can result in a very wide column, which isn't useful.  :grin:

stevesims:
Fixes #432

Ensures that fixed width `GridCol` elements get `flex-shrink: 0` set

**Checklist**:

* [ ] Documentation  N/A
* [ ] Tests  N/A
* [x] Ready to be merged

### DIFF
--- a/components/grid-col/src/__snapshots__/test.js.snap
+++ b/components/grid-col/src/__snapshots__/test.js.snap
@@ -8,16 +8,16 @@ exports[`GridCol matches snapshot: enzyme.mount 1`] = `
 
 @media only screen and (min-width:641px) {
   .emotion-0 {
-    margin-right: 15px;
-    margin-left: 15px;
+    padding-right: 15px;
+    padding-left: 15px;
   }
 
   .emotion-0:first-child {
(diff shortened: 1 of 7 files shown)


penx:
I think Chromatic is showing some unintended changes

https://www.chromaticqa.com/snapshot?appId=5ad61aab7f00090020dfdba8&id=5bd6f58ac6f51b0024da427c

PR #451, opened 2018-11-05
Page Component

### LINKED ISSUE #426: Page Component
Create a page component, mirroring govuk-frontend's page templates and layout documentation.

https://design-system.service.gov.uk/styles/page-template/
https://design-system.service.gov.uk/styles/layout/#page-wrappers

- Follow the structure of the nunjucks templates
- Don't include html tag and head tag content
- Use variable and block names from the nunjucks templates for the props used by the React version, other than where changing a prop name to 'children' would be more appropriate for React.
- Use CSS class names as guidance for component/atom names 

Should use atoms tracking [MainWrapper](https://github.com/alphagov/govuk-frontend/blob/master/src/objects/_main-wrapper.scss) and [WidthContainer](https://github.com/alphagov/govuk-frontend/blob/master/src/objects/_width-container.scss) (was #427)

Proposed usage:

```jsx
import React from 'react';
import { Page, TopNav, BackLink, H1, Footer } from 'govuk-react';

const MyPage = () => <Page
  header={<TopNav />}
  footer={<Footer />}
>
  <React.Fragment>
    <BackLink />
    <Page.Main>
      <H1>Page Title</H1>
    </Page.Main>
  </React.Fragment>
</Page>
```

If header or footer are not set, a default should be rendered using defaultProps.

penx:
Fixes #426 

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged

Further stories and tests in new ticket to be followed up in #456

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -26,7 +26,10 @@ module.exports = {
         "import/no-extraneous-dependencies": ["error", {"devDependencies": true}]
       },
       "settings": {
-        "import/core-modules": Object.keys(devDependencies)
+        "import/core-modules": [
+          'govuk-react',
+          ...Object.keys(devDependencies)
+        ]
       }
     },
     {
--- a/components/page/src/index.js
+++ b/components/page/src/index.js
@@ -6,14 +6,13 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
-import styled from 'react-emotion';
 import TopNav from '@govuk-react/top-nav';
 // import Footer from '@govuk-react/footer';
 
 import Main from './atoms/main';
 import WidthContainer from './atoms/width-container';
 
-const PageWrapper = styled('div')();
+// const PageWrapper = styled('div')();
 
 // TODO: replace with actual footer component once built
 // const Footer = styled('footer')({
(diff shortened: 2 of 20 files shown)


penx:

Built with commit fdde922ab4b14dd5daecdb62736b5091c7bc428b

https://app.netlify.com/sites/affectionate-nobel-f3a8ec/deploys/5be05b5982d3f1196f3d0c38

.eslintrc.js
```diff
@@ -26,7 +26,10 @@ module.exports = {
         "import/no-extraneous-dependencies": ["error", {"devDependencies": true}]
       },
       "settings": {
-        "import/core-modules": Object.keys(devDependencies)
+        "import/core-modules": [
+          'govuk-react',
```
stevesims: curious as to what this change does :grin:

penx: https://github.com/UKHomeOffice/govuk-react/pull/451/files#diff-77150f3501b5f8d820e6f61a753cb5f9R2

I'm importing from govuk-react in a fixture rather than importing from the specific sub packages. Because it's easier and feels better.

This was coming up with an eslint error as govuk-react wasn't included in the devDependencies. I'm adding it as an exception in our eslint rules as with all the other monorepo packages.

components/page/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+
+Page.defaultProps = {
+  children: undefined,
+  header: <TopNav />,
+  footer: <Footer />,
+};
+
+Page.Main = Main;
+// if WidthContainer is exported, should be combined with a use case and story
+// Page.WidthContainer = WidthContainer;
```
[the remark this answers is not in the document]
penx: That's pretty much the only functionality Page is doing right now, in addition to the child Main element that is exported.

I would suggest that if they don't want this feature then they shouldn't be using the Page component.

I'm also hesitant to add speculative features at this stage.

penx:
This is a purely visual component and has 100% coverage through snapshots so marking the tests as done.

.eslintrc.js
stevesims: ah ok - cool :grin:

penx:
Not sure why Chromatic is not displaying the full component, but it's appearing fine in storybook so have approved

PR #457, opened 2018-11-06
Remove margin-bottom from grid components

jsrobertson:
* Remove margin-bottom from both grid-row and grid-col as per:
https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_grid_layout.scss#L66
https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_grid_layout.scss#L93

**Checklist**:
* [x] Documentation
* [x] Tests
* [x] Ready to be merged

### DIFF
--- a/components/grid-col/src/__snapshots__/test.js.snap
+++ b/components/grid-col/src/__snapshots__/test.js.snap
@@ -2,7 +2,6 @@
 
 exports[`GridCol matches snapshot: enzyme.mount 1`] = `
 .emotion-0 {
-  margin-bottom: 15px;
   box-sizing: border-box;
 }
 
--- a/components/grid-col/src/__snapshots__/test.js.snap
+++ b/components/grid-col/src/__snapshots__/test.js.snap
@@ -3,13 +3,8 @@
 exports[`GridCol matches snapshot: enzyme.mount 1`] = `
 .emotion-0 {
   box-sizing: border-box;
-}
-
-@media only screen and (min-width:641px) {
-  .emotion-0 {
-    padding-right: 15px;
-    padding-left: 15px;
-  }
+  padding-right: 15px;
+  padding-left: 15px;
 }
 
 @media only screen and (min-width:641px) {
--- a/components/grid-col/src/index.js
+++ b/components/grid-col/src/index.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import styled from 'react-emotion';
 import PropTypes from 'prop-types';
-import { GUTTER_HALF, MEDIA_QUERIES, SPACING } from '@govuk-react/constants';
+import { GUTTER_HALF, MEDIA_QUERIES } from '@govuk-react/constants';
 
 const colValues = {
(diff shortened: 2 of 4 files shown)


penx:

Built with commit 3d7f2c0b48bd84ca39f07d571f9403eb840d4231

https://app.netlify.com/sites/affectionate-nobel-f3a8ec/deploys/5be212fdc965926d94c81323

penx:
We should be tracking govuk-frontend rather than govuk_frontend_toolkit if possible. 

https://design-system.service.gov.uk/styles/layout/#grid-system

From what I can tell neither have margin bottom, but I need to double check.

I wonder if this was added for any reason.. hopefully will be picked up by visual regression if so.

This may need a 0.4 release.

penx:
@stevesims changed the grid 8 days ago so should take a look https://github.com/UKHomeOffice/govuk-react/pull/443

stevesims:
The grid components were originally written to follow the patterns from the old govuk-elements, which used padding rather than margins, and had a padding-bottom set.

When I recently converted them to work based on margins I left the bottom padding/margin in place.  So yeah, that's an error, as it's not following the proper patterns.  My bad!

I wonder tho if these components should get `withWhiteSpace` applied to them (with a zero margin set) which would allow users of them to apply bottom margins if desired.  We may want/need this for some places where we're using these components.

penx:
Yep I think they should have withWhitespace but doesn't have to be part of this PR.

We really need to sort out our CI failing when PRs are from external forks.

jsrobertson:
@stevesims @penx I've just added another commit which maintains the column padding at smaller screen sizes. As per, https://github.com/alphagov/govuk-frontend/blob/ad596de756ef51619c6e5bc8f22f3b93479773dd/src/helpers/_grid.scss#L93

penx:
cheers, are you in a rush to see this merged and released? I want to try get the CI fixed so that it doesn't error on external PRs. May look in to that now...

penx:
@stevesims it's ok I know what the issue with env variables I just haven't figured out what to do about it yet.

PRs from external sources don't have access to secret env variables as that would create a security hole.

penx:
https://github.com/UKHomeOffice/govuk-react/pull/464 should resolve PR issue

jsrobertson:
@penx No major rush, I can get by for now but would be helpful to get using it. Happy to help with publishing etc if needs be, just give me a shout. Cheers 👍

penx:
I rebased off master. CI is building now but failing due to a snapshot not matching

jsrobertson:
Apologies. I've updated the snapshot now and pushed.

penx:
This seems a bit odd https://www.chromaticqa.com/snapshot?appId=5ad61aab7f00090020dfdba8&id=5be852f948642300243b4676

penx:
Ah all good. The issue in Chromatic is because the Chromatic UI is vertically centering the before/after images

penx:
merged, thanks @jsrobertson!

PR #486, opened 2019-01-07
Change Jest version

gavinorland:
This updates to an alpha version of Jest which supports the `babel.config.js` file format being used since [this commit](https://github.com/govuk-react/govuk-react/commit/9b42259f8db13f8c1403a415d01f8a6f53795725).

**Checklist**:

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
       "!<rootDir>/packages/api-docs/**",
       "!<rootDir>/packages/storybook-components/**"
     ],
-    "setupTestFrameworkScriptFile": "<rootDir>/scripts/setupTests.js",
+    "setupFilesAfterEnv": ["<rootDir>/scripts/setupTests.js"],
     "testMatch": [
       "<rootDir>/{components,packages}/*/src/**/?(*.)(spec|test).js?(x)"
     ],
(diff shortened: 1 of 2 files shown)


penx:

Built with commit 720765f57d39aade9f2bfabcc5763cb6bebe67da

https://app.netlify.com/sites/govuk-react/deploys/5c3373c5d9c9f4000aed46b9

penx:
What was broken that requires this change to be in now?

Why should we use the alpha rather than waiting for release?

Do you have a link to any related Jest issues describing the upcoming feature we need?

gavinorland:
None of the tests can be run with the code as it currently is. https://github.com/facebook/jest/issues/6908 I guess we could revert to the previous Babel file format (I'm not sure) but otherwise I think we will need to move to at least this alpha version.

penx:
Our tests run on the CI so I don’t understand how they can be failing. I’ll take a closer look Wednesday

gavinorland:
(Just trying to see why build is failing)

gavinorland:
Hmm, the build is failing at the moment with:
```
TypeError: Cannot read property 'concat' of undefined
    at Object.<anonymous> (/home/travis/build/govuk-react/govuk-react/node_modules/jest-resolve/build/isBuiltinModule.js:29:27)
    ...
```
I can't see the cause of this so far... I'll leave it for the time being and we could pursue it if @penx also finds this PR to be necessary.

PR #487, opened 2019-01-07
Remove navigator check

gavinorland:
The check inside this component is not required as it's made [elsewhere](https://github.com/govuk-react/govuk-react/blob/master/packages/storybook/src/stories/index.js#L8) at a higher level.

**Checklist**:

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged


penx:

Built with commit 8cfc0c153747952149032eca69071cc3be31b3cd

https://app.netlify.com/sites/govuk-react/deploys/5c3386e1d9c9f4000aed6476

penx:
Is there a benefit to removing this? It the example usage it is not needed, but `storybook-components` is exported so could potentially be used in other places. I don't see any harm in keeping the check in both places.

Issue #461, opened 2018-11-08
Radio buttons and checkboxes have no background


stevesims:
When using the radio or checkbox components on a non-white page you'll see that the radio and checkbox are transparent.

To be consistent with other input controls they should get a white background-color set.

penx:
Some progress was made on https://github.com/govuk-react/govuk-react/pull/488 but work has halted and PR has been closed for now. Please reference comments there.

stevesims:
We need to create an issue on govuk-frontend for this, showing an example of this, as our current styling matches theirs.

Meanwhile we can add support for setting the background colour in our own component, in much the same way that we now allow custom colours in our button component - have the default set to `transparent` to match govuk-frontend.

nalanirojas25:
I can attempt to fix this issue. However, I'm having trouble loading the page in the browser. npm start isn't working. What command is used to bring up the content in the browser?

penx:
@jsrobertson said he could look in to it 2 years ago... 😄

https://github.com/govuk-react/govuk-react/pull/488#issuecomment-455185118

PR #488, opened 2019-01-08
Set component backgrounds to white

gavinorland:
Addresses #461.

**Checklist**:

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged

### DIFF
--- a/components/checkbox/src/__snapshots__/test.js.snap
+++ b/components/checkbox/src/__snapshots__/test.js.snap
@@ -73,7 +73,7 @@ exports[`Checkbox matches wrapper snapshot: wrapper mount 1`] = `
   content: '';
   display: block;
   border: 2px solid #0b0c0c;
-  background: transparent;
+  background: white;
   width: 34px;
   height: 34px;
   position: absolute;
@@ -177,7 +177,7 @@ exports[`Checkbox renders disabled checkbox: disabled 1`] = `
   content: '';
   display: block;
   border: 2px solid #0b0c0c;
-  background: transparent;
+  background: white;
   width: 34px;
   height: 34px;
   position: absolute;
--- a/components/checkbox/src/index.js
+++ b/components/checkbox/src/index.js
@@ -61,7 +61,7 @@ const StyledLabel = styled('span')({
     content: "''",
     display: 'block',
     border: `2px solid ${BLACK}`,
-    background: 'transparent',
+    background: 'white',
     width: '34px',
     height: '34px',
     position: 'absolute',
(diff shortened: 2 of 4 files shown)


penx:

Built with commit 18beca648d2ca490542c7d0fb19c513c49047db0

https://app.netlify.com/sites/govuk-react/deploys/5c348a842b93c60009909188

penx:
I don't understand the underlying issue:

- unless there is a clear need to divert, we should be mirroring govuk-frontend sass
- it's not intended to use these components on a non white background
- if an application wants to have custom styles, we can make some adjustments to support this, but I am not convinced that setting the background to always be white would help more than hinder custom styles
- is there a screenshot of the current issue?
- can we reflect the issue this is solving in a story in storybook?

stevesims:
In our app we have a light grey background - as checkboxes and radio buttons have no background color set, whereas other `input` elements *do* have a white background color set, they look out of place and wrong.  They are inconsistent with their peers.
This can be seen here: https://codesandbox.io/s/wojk6jwpzl

We could allow a prop to set this color as an alternative.  Re-styling is not an option given how these elements are produced and styled.

The same bug exists in govuk-frontend.  I have not yet filed an issue against them, or raised a PR, mostly because creating an example would take time that I've not had so far.

penx:
ok cool thanks for the demo

- what does govuk-frontend do? Does this issue also exist in govuk-frontend?
- if we do this, I would prefer it was a prop that defaults to white but can be overridden
- there should be a story in storybook showing this (e.g. so it is picked up if it is broken in future)

jsrobertson:
Hi @stevesims @penx.  Just come across this while implementing govuk-react checkbox and radio into our apps.

We also have the checkbox on some non white backgrounds (rightly or wrongly) and have stumbled across this issue.  For now though I've targeted in a hacky CSS file we have to get round the issue.

I agree with @stevesims on:
> if we do this, I would prefer it was a prop that defaults to white but can be overridden

I have also noticed that I think the styles on both the radio and checkbox components are slightly different to the latest govuk-frontend ones so I'm wondering if it's best to have an issue.  Checkboxes are now 40px instead of 38px for instance.

stevesims:
We have some drift from govuk-frontend in all our input stylings, mostly because the majority of these components were produced before govuk-frontend was released and thus were based on earlier GDS styles.

I started to work this week on attempting to address this for the `Input` component, but fell down a rabbit-hole related to typography, which I hope to manage to crawl out of tomorrow.  :grin:

Unfortunately @gavinorland  isn't available to finish off addressing @penx's feedback on this issue right now, so if you fancy tackling it @jsrobertson that would be great and we'd welcome either a different PR or further commits against this PR.

penx:
The intention is to completely match govuk-frontend though, so where we don't match then yes issues/bugs can be raised.

nalanirojas25:
@jsrobertson are you actively working on this issue? I was interested in giving it a shot.

Issue #449, opened 2018-11-05
BackLink should only underline if href is provided


penx:
BackLink should only underline is href is provided (or if an onClick is provided, which in turn should set the href to # if it has not been set).

This is to mirror default browser handling of anchors that are missing an href.

penx:
We can either revisit https://github.com/govuk-react/govuk-react/pull/489 or create a new branch to do this

penx:
we should also be mirroring what govuk-frontend does so if this is also a bug with govuk-frontend we should raise it there

penx:
Raised on govuk-frontend. As this library intends to mirror govuk-frontend, I think we should only address this issue if it's also addressed on govuk-frontend https://github.com/alphagov/govuk-frontend/issues/1171

penx:
This was fixed on govuk-frontend so we should also fix here

PR #489, opened 2019-01-08
BackLink - Set href to # if onClick provided but no href

gavinorland:
This commit addresses #449.

- If an `onClick` is provided but no `href` is provided, the `href` is set to be "#".
- Stories updated
- Tests amended
-- shallow used
-- Sinon replaced with `jest.fn()`
-- New tests added

**Checklist**:

* [ ] Documentation N/A
* [x] Tests
* [x] Ready to be merged

### DIFF
--- a/components/back-link/src/index.js
+++ b/components/back-link/src/index.js
@@ -10,7 +10,9 @@ import {
   NTA_LIGHT,
 } from '@govuk-react/constants';
 
-const Anchor = styled('a')({
+const Anchor = styled('a')(({
+  href, onClick,
+}) => ({
   fontFamily: NTA_LIGHT,
   WebkitFontSmoothing: 'antialiased',
   MozOsxFontSmoothing: 'grayscale',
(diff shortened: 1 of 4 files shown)


penx:

Built with commit 9c495512d14e64004197e069116abaac2f625074

https://app.netlify.com/sites/govuk-react/deploys/5c34e1b73a94a50008810bca

components/back-link/src/index.js
```diff
@@ -28,7 +30,7 @@ const Anchor = styled('a')({
   border: 0,
   backgroundColor: 'transparent',
   color: `${BLACK}`,
-  borderBottom: `1px solid ${BLACK}`,
+  borderBottom: href ? `1px solid ${BLACK}` : 'none',
```
penx: minor but I think I'd rather this was done in the same way as browser default stylesheets, we shouldn't need to compute an additional class:

The old way:
```css
a[href] {
border-bottom: none
}
```

The new way (maybe best avoid for now as it's experimental):
https://developer.mozilla.org/en-US/docs/Web/CSS/:any-link

penx:
approved other than minor comment

gavinorland:
@penx I'm seeing a strange problem with this whereby the snapshot generated by Travis differs from the snapshot generated locally. Perhaps this is due to me having to use a later version of Jest in order to run tests at all, as per https://github.com/govuk-react/govuk-react/pull/486. Maybe you could pull the branch and see how you get on.

penx:
There's a few conflicts and @stevesims is concerned the PR may be obsolete given styling changes - either way it would likely be faster to reimplement this at a later date using this PR as a guide

PR #512, opened 2019-01-24
Update Button styling

stevesims:
`Button` component now styled more consistently with govuk-frontend
Supports following additional new props:
* `buttonColour` to override default button colour
* `buttonHoverColour` which defaults to `buttonColour` darkened by 5%
* `buttonShadowColour` which defaults to `buttonColour` darkened by 15%
* `buttonTextColour` which defaults to govuk white

TODO:
`withWhiteSpace` HOC does not yet support `adjustment` value, which `Button` now attempts to use, so this needs to be added - however that can be done in a later PR

**Checklist**:

* [x] Documentation
* [ ] Tests
* [x] Ready to be merged

### DIFF
--- a/components/button/src/index.js
+++ b/components/button/src/index.js
@@ -114,7 +114,9 @@ const StyledButton = styled('button')(
 
     ' svg': {
       maxWidth: '15px',
-      marginLeft: SPACING_POINTS[4],
+      [MEDIA_QUERIES.TABLET]: {
+        marginLeft: SPACING_POINTS[2],
+      },
     },
   }),
 
@@ -131,6 +133,10 @@ const StyledButton = styled('button')(
     return undefined;
   },
 );
+
+const ButtonContents = styled('span')({
+  flexGrow: 1,
+});
 
 /**
  *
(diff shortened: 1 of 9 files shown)


penx:

Building with commit 75dd871d55edcaf30acf85b760e55695696f345f

https://app.netlify.com/sites/govuk-react/deploys/5c49e9a5234e5f000872fb44

penx:
Where's the code that ensures buttonHoverColour is only calculated if the prop isn't set? It looks like this is working but I don't see how :-)

components/button/src/index.js
```diff
@@ -2,69 +2,142 @@ import styled from 'react-emotion';
...
+  ({ isStart }) => typography.font({
+    size: isStart ? 24 : 19,
+    lineHeight: isStart ? '1' : '19px',
+    weight: isStart ? 'bold' : undefined,
+  }),
+  FOCUSABLE,
+
+  ({
+    buttonColour = BUTTON_COLOUR,
+    buttonHoverColour = darken(0.05, buttonColour),
```
stevesims: This is where button colours get calculated, if a prop isn't provided to override.

penx: ah yeah this is the default value, was reading it as an assignment

Issue #286, opened 2018-05-24
Consider `mb` vs `marginBottom` in `withWhiteSpace` HOC


stevesims:
the `withWhiteSpace` HOC uses both `mb` and `marginBottom` to set the `marginBottom` style, which is potentially slightly confusing.  we should consider making this less confusing somehow ;)

stevesims:
IIRC there may be a difference between the two - `mb` doesn't directly set the `marginBottom` value, but rather takes a value that's an index into a set of different sizes

`mb` is a prop that can be set on an element at display-time

`marginBottom` is the default value that gets set when configuring the `withWhiteSpace` HOC (applying it to another component)

Loque-:
I have just used `mb=""` in an example and I personally would prefer a full-name over an abbreviation to keep with convention (most React things I have been exposed to are explicit and not abbreviated) and I think this change would help new users understand what it is.

As an initial idea something like `spacingBottom` could be nice(?) to directly replace the names `mb` and `marginBottom`, as it is spacing that follows a scale, and would avoid any confusion with the CSS value margin-bottom that could have it's own px/em/% value (despite it using margin to render the spacing value).

penx:
For reference, mb convention comes from grid-styled

> I had mentioned using mb as the prop name as this is the naming convention used by https://github.com/jxnblk/grid-styled
>
> Perhaps in a comment above StyledHoc.propTypes = { we could mention that this is the naming convention we're using plus link?


https://github.com/govuk-react/govuk-react/pull/185#discussion_r179383795

PR #515, opened 2019-01-29
Enhanced withWhiteSpace HOC

stevesims:
* `withWhiteSpace` HOC updated
  - supports `margin` and `padding` props and config
  - values can be:
    - single number (size to use in all directions)
    - object in format `{ size, direction, adjustment }` (as per `responsivePadding` etc lib call)
    - array of numbers/objects
  - `mb`/`marginBottom` still supported for backward compatibility
  - example stories cleaned up
    - no need to double-wrap `withWhiteSpace`
    - examples for new options added against a demo component
* `spacing` lib enhanced
  - styles no longer set as `px` strings, as emotion does that automatically
  - `responsiveMargin` and `responsivePadding` now support single integer size value
* fix: `InputField` no longer double-wraps `withWhiteSpace`

**Checklist**:

* [ ] Documentation
* [x] Tests
* [x] Ready to be merged

### DIFF
--- a/components/error-summary/src/__snapshots__/test.js.snap
+++ b/components/error-summary/src/__snapshots__/test.js.snap
@@ -245,24 +245,7 @@ exports[`error summary matches the ErrorSummary snapshot: error summary 1`] = `
   }
 }
 
-.emotion-82 {
-  margin-bottom: 0;
-  margin-bottom: 0;
-}
-
-@media only screen and (min-width:641px) {
-  .emotion-82 {
-    margin-bottom: 0;
-  }
-}
-
-@media only screen and (min-width:641px) {
-  .emotion-82 {
-    margin-bottom: 0;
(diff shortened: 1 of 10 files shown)


penx:
This opens an interesting discussion :-) 

the `mb` convention I had copied from grid-styled, which is now part of rebass
https://github.com/govuk-react/govuk-react/issues/286

A few things have directed me previously to consider having govuk-react use rebass for some utility functions.

On the other hand, maybe we should be mirroring govuk-frontend's utility functions and naming conventions.

Happy for this to be merged as is but wanted to raise that we're heading in one direction over the other (and we may want to revisit this decision at a later date!)

penx:
As @stevesims mentioned, it's easier for us to convert/track govuk-frontend sass if we're also converting their util functions

PR #521, opened 2019-01-30
Replace emotion with styled-components

### LINKED ISSUE #485: Upgrade to emotion 10
We've been using this package for a bit over here in Canada 👍. Any thoughts about upgrading `emotion` to version 10?

### LINKED ISSUE #520: Consider switching to styled-components
Recently we've tried to put some work into upgrading to Emotion 10 as suggested in #485.  Unfortunately for a component library such as this, the upgrade path is not clean and it would be a "big bang" update, dropping backward compatibility to Emotion 9.  Whilst progress was made towards this in #491 some significant difficulties were encountered and work has currently stalled.

More recently during work to improve our compatibility with govuk-frontend, some serious problems were encountered when attempting to replicate the styling of list components.  In brief, there is a style rule to ensure that lists placed within lists get a `margin-top: 10px` style set - in principle this is a simple rule to implement however, long story short, we could not get it to work reliably/properly using Emotion.

Part of the problem encountered is illustrated in [this sandbox example](https://codesandbox.io/s/k08pmjrl5r).  (NB this example uses Emotion 10, but the same issue occurs in Emotion 9.)  The problems however were wider and deeper than this simple example, with additional issues caused by wrapping our components with the `withWhiteSpace` HOC.

Preliminary investigations seem to indicate these problems do not exist with styled-components, as can be seen by changing the above sandbox example to use styled-components.

As moving to Emotion 10 is proving problematic, and we are finding additional issues with Emotion, we should consider switching to styled-components

stevesims:
Resolves #520, closes #485

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged


penx:
I was gonna check keyframes, but the loadingbox has been broken for some time, so not much to check

https://5c52e3939f97fe000865d6c2--govuk-react.netlify.com/?selectedKind=Misc%2FLoadingBox&selectedStory=Component%20default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

(doesn't fade in/out)

penx:
I don't think there's much to check here as long as the chromatic specs remain unchanged

Issue #522, opened 2019-01-31
Styling issues caused by withWhiteSpace HOC


stevesims:
In attempting to update the styles for list components to better match govuk-frontend, it's been discovered that using the `withWhiteSpace` HOC can be problematic, causing unintended consequences when attempting to set styles for child components.

Take this example of a very simplified version of the `OrderedList` component:
```
const OrderedList = styled('ol')(
  {
    marginTop: 0,
    // Lists within a list should get a marginTop
    [`& &, ${UnorderedList}`]: {
      marginTop: 10,
    },
  }
)
```

This component forms part of a family of 3 list components together with `UnorderedList` and `ListItem`.

When using Emotion with the above code, `ListItem` components get a `marginTop: 10` set.  The reason for this is that as all 3 list components are wrapped with the `withWhiteSpace` HOC, the class name value for the `UnorderedList` selector above is identical to a class name being applied to `ListItem` in the DOM - it is effectively a `withWhiteSpace` selector.

With styled-components, the logical way to implement `UnorderedList` is to just define it as `OrderedList.withComponent('ul')`, and thus the selector in the example above becomes just `& &`.  Unfortunately when using nested `UnorderedList`, whilst the `marginTop` rule ends up being applied, but other styles for `UnorderedList` do *not*.  This again appears to be down to wrapping with the `withWhiteSpace` HOC.

In both cases the simple fix is to remove the `withWhiteSpace` HOC wrapping and just use the styled components directly.  This is not desirable however as it removes support for adjusting margins and padding.

It is thus suggested that the `withWhiteSpace` HOC be deprecated/retired, and replaced with a styling function generator that can provide the same support.

This would be used thus:
```
const Component = styled('div')(
  { ...some styles here },
  withWhiteSpace(configAsPerExistingHOC)
);
```

PR #523, opened 2019-01-31
Add `withWhiteSpace` to spacing lib

stevesims:
* adds new `withWhiteSpace` function to spacing lib
  - generates spacing styles, accepting `margin`, `padding` and `mb` props
  - intended for use within styled component definitions

`withWhiteSpace` HOC refactored to use function, but is now deprecated

Aims to provide a solution to #522

**Checklist**:

* [ ] Documentation
* [x] Tests
* [x] Ready to be merged

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -83,6 +83,15 @@ We want to be router agnostic. We want to support parent projects using React Ro
 
 ### White Space
 
-Components are built to have no white space around them, and are then wrapped with the withWhiteSpace HOC where we can provide some default values. This allows the parent application to override the defaults with e.g. an `mb` prop.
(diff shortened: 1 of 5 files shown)


penx:
The CONTRIBUTING.md doc will need updating 
https://github.com/govuk-react/govuk-react/blob/master/CONTRIBUTING.md#white-space

penx:
ah ok maybe not breaking... but breaking when we move to styled-components?

stevesims:
Yeah - this won't be breaking by itself.  When we actually remove the `withWhiteSpace` HOC that'd be breaking.  And yes - moving to styled-components changes the peer dependency, so that'd breaking too.

PR #546, opened 2019-02-11
Add new Link component

stevesims:
New `Link` component added

Updated `asAnchor` HOC to use it, but also deprecate `asAnchor` as use of styled-components now means that the `Link` component can re-style router links.

Documentation, tests, and examples all updated to no longer use `asAnchor`

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged


penx:

Building with commit 3f09db3a364cd8c293acc30e9ca5bd5fff68d192

https://app.netlify.com/sites/govuk-react/deploys/5c6183f62f0b970007533e99

penx:
Why are there changes showing in storybook? PR description doesn't seem to imply any visual change.

penx:
btw, it may be a good idea to expand on our PR process for storybook:

Previously (informally):

> if the raiser of the PR is a maintainer, they can accept storybook changes as they are best placed to distinguish between intended and unintended changes

Proposed:

> if the raiser of the PR is a maintainer, they can accept storybook changes as they are best placed to distinguish between intended and unintended changes, *as long as the PR description contains screenshots/diffs giving a good summary of those changes*

(this should just be a case of copying one or two of the storybook image urls and pasting in to PR description when you accept them)

What do you reckon?

components/breadcrumb/src/index.js
```diff
@@ -63,6 +63,7 @@ const BreadcrumbListItem = styled('li')({
   },
 
   // Additions to help ensure links get coloured as we expect
+  // NB specificity of this should override default in Link component
```
penx: 😢 is this from govuk-frontend?

stevesims: hmmm - i had thought it was, but actually it isn't

on a quick look just now, it seems that breadcrumb doesn't work quite the same.  might need some more looking into to work out what exactly should be going on here

stevesims:
The storybook change i believe is down to some slight differences in how the old `asAnchor` styled links vs the new Link component

As for the visual diff thing, perhaps a simple description would suffice - otherwise in some cases we could end up with very long PRs that take a long time to document?

stevesims:
I've looked deeper into this now, and the storybook change for the error-summary component is down to changes between the style the previous `asAnchor` HOC added and the base appearance of the `Link` component.  It brings the appearance of our error-summary to be slightly closer to govuk-frontend, however there are still differences which need to be addressed.

components/breadcrumb/src/index.js
stevesims: having looked a bit further, our breadcrumb component differs slightly from govuk-frontend, i believe to make it "more reacty"...

the underlying issue here is with how the link should get styled.  we had given examples that used `asAnchor`, which have now been changed to use `Link`.  a better solution would be to provide a `Breadcrumb.Link` component/atom that provides styles specific to the breadcrumb system, and for the examples to be changed to use that

stevesims: ok - styling bodge removed - new `Breadcrumb.Link` atom added and docs/examples changed accordingly

stevesims:
The remaining link/anchor styling work to be done from this point is in how links are handled in the `top-nav` component.

At present, govuk-react does not include support for indicating "active" nav links, however the [`page-header` component of govuk-frederic](https://github.com/stevesims/govuk-frederic/tree/master/components/page-header) does.

I'd be inclined at this point to merge this as-is, and work on updating `top-nav` in a separate PR

penx:
As discussed:

- we don't currently have any examples of links with active (class not pseudo) styles so don't have any stories showing compatibility with Next/React/Reach router
- we have agreed that when we do support active classes, we will  support this being optionally provided as an active prop or an active CSS class with the CSS class defaulting to active and being customisable via a prop called activeClassName.
-The active Class will always be appended to the generated class e.g. .generated.active, so we can be sure there would not be conflicts e.g. with cascade

PR #560, opened 2019-02-20
use spacing and typography lib

### LINKED ISSUE #522: Styling issues caused by withWhiteSpace HOC

stevesims:
most `withWhiteSpace` HOC wrapping now done with spacing lib

bug fixes to styling in `Details` component
  * marginBottom wasn't configured correctly
  * `:hover` had extraneous `:` preventing it from working

Some remaining font styling replaced with calls to typography.font

Fixes #522

**Checklist**:

* [x] Documentation
* [ ] Tests N/A
* [x] Ready to be merged


penx:

Building with commit cf736749947df3f1ea795fa8da9f66c4ba75817c

https://app.netlify.com/sites/govuk-react/deploys/5c6d47ae3a525900098d7f75

penx:

Building with commit 2a475be00707acde7b273dccf1c2c484bcaa3830

https://app.netlify.com/sites/govuk-react/deploys/5c6d47f3c3a720000a082105

stevesims:
Improved standardisation of whitespace and typography has meant lots of small visual diffs, mostly with fixed line-spacing

components/details/src/index.js
```diff
@@ -92,8 +92,8 @@ const DetailsText = styled('div')({
  * - https://github.com/alphagov/govuk-frontend/blob/master/src/components/details/_details.scss
  */
 const Details = ({ summary, children, ...props }) => (
-  <StyledDetails {...props}>
-    <StyledSummary><SummaryText>{summary}</SummaryText></StyledSummary>
+  <StyledDetails {...props} role="group">
```
penx: please review this bit

PR #563, opened 2019-02-22
Width settings

stevesims:
* `WIDTHS` constant object added
* `withWidth` added to spacing lib
  - produces a responsive width style
    - 100% width, with specified width for tablet or greater
    - no style if no config/prop provided
  - accepts a `width` config value
  - accepts a `setWidth` prop
  - config/prop accepts key from `WIDTHS` config, or any value
  - config also supports `mediaQuery` (to pick a different query) and `noDefault` (to omit 100% width)
* `Table.Cell` now uses `withWidth`
  - `Table` examples and docs updated to demo use
* `GridCol` updated
  - `columnThreeQuarters` and `columnFull` added
  - `column*` props however now deprecated in favour of...
  - `setWidth` prop, provided with the aid of `spacing.withWidth`
  - `setDesktopWidth` prop allows desktop breakpoints for column widths

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged

### DIFF
--- a/packages/constants/src/spacing.js
+++ b/packages/constants/src/spacing.js
@@ -77,3 +77,12 @@ export const SPACING_POINTS = {
   8: 50,
   9: 60,
 };
+
+export const WIDTHS = {
+  'one-quarter': '25%',
+  'one-third': '33.3333%',
+  'one-half': '50%',
+  'two-thirds': '66.6666%',
+  'three-quarters': '75%',
+  full: '100%',
+};
--- a/packages/constants/src/spacing.js
+++ b/packages/constants/src/spacing.js
@@ -78,6 +78,7 @@ export const SPACING_POINTS = {
   9: 60,
 };
 
+// Ref: https://github.com/alphagov/govuk-frontend/blob/68bd09bb3e54b7ef4b4084ad2b3336858923a041/src/settings/_measurements.scss#L23
 export const WIDTHS = {
   'one-quarter': '25%',
   'one-third': '33.3333%',
--- a/API.md
+++ b/API.md
@@ -539,9 +539,11 @@ import GridCol from '@govuk-react/grid-col';
 Prop | Required | Default | Type | Description
 :--- | :------- | :------ | :--- | :----------
  `children` |  | ```undefined``` | node | GridCol content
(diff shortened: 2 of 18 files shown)


penx:

Building with commit cb7ec03a056200a8675499e0c66f6e9766491b58

https://app.netlify.com/sites/govuk-react/deploys/5c6fc4e868a9e600085899e1

stevesims:
This PR currently only adds custom width setting to `Table.Cell`.  In govuk-frontend however as this feature is provided via override classes which can be applied to every element/component we need to spread this support wider.

Rather than applying `withWidth` to all components, we will likely be a bit more selective, as for some components it's less applicable/desirable to offer this facility.  This will thus be added on a case-by-case basis, either in subsequent commits to this PR or on later PRs

stevesims:
Updated with further enhancements to GridCol, which can now support all styling options of the govuk-frontend grid (plus more)

packages/constants/src/spacing.js
```diff
@@ -77,3 +77,12 @@ export const SPACING_POINTS = {
   8: 50,
   9: 60,
 };
+
+export const WIDTHS = {
+  'one-quarter': '25%',
+  'one-third': '33.3333%',
+  'one-half': '50%',
```
penx: `one-` seems redundant and only seems to make code more verbose than necessary, especially in `one-half`. Do these names come from somewhere? Also 🚲

stevesims: they come from govuk-frontend
https://github.com/alphagov/govuk-frontend/blob/68bd09bb3e54b7ef4b4084ad2b3336858923a041/src/settings/_measurements.scss#L23

Having said that tho - those measurements settings are what's used for grid, rather than width.  You're right - they're redundant - and can maybe be got rid of, or supplemented with versions that don't include "one"

🚲 indeed :wink:

stevesims: Will add in reference to measurements link into the code

penx:
as discussed, please add a story that shows how Table works on mobile, given I think width:100% is being set on all table cells below tablet.

PR #571, opened 2019-02-26
Add Heading component to replace header

### LINKED ISSUE #454: Header should be called Heading
To be in line with govuk-frontend, Header should refer to the Top Nav:

https://design-system.service.gov.uk/components/header/

We can use Heading to refer to our title/H1 components instead, freeing the name 'Header' up for using for the top nav in future.

Once Header is renaming Heading, we should mark Header as deprecated.
(as long as we can un deprecate it in future)

stevesims:
Heading component allows Header to be deprecated.

Deprecation currently marked in documentation only.

Closes #454 

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged

### DIFF
--- a/components/caption/src/index.js
+++ b/components/caption/src/index.js
@@ -29,6 +29,7 @@ const StyledCaption = styled('span')(
     };
   },
   {
+    display: 'block',
     color: SECONDARY_TEXT_COLOUR,
   },
   spacing.withWhiteSpace(),
@@ -45,7 +46,7 @@ const StyledCaption = styled('span')(
  *
  * With another header
  * ```jsx
- * import { H1 } from '@govuk-react/header';
+ * import { H1 } from '@govuk-react/heading';
  *
  * <Caption size="XL">Supporting header text</Caption>
  * <H1>Main header text</H1>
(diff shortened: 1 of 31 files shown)


penx:

Building with commit 10ea165690c5a7b600a1fd36b66da18e30235ad1

https://app.netlify.com/sites/govuk-react/deploys/5c751ec6a097d8000771dc31

components/caption/src/index.js
```diff
@@ -29,6 +29,7 @@ const StyledCaption = styled('span')(
     };
   },
   {
+    display: 'block',
```
penx: why was this added?

stevesims: it's prior omission was a bug

components/heading/src/index.js
```diff
@@ -0,0 +1,120 @@
...
+import { spacing, typography } from '@govuk-react/lib';
+
+// use `size` only with string for XLARGE, SMALL etc and number for px size
+// so if `size` is a string, we find a numeric size based off `HEADING_SIZES`
+// but if `size` is a number we just send through that number
+
+const StyledHeading = styled(({
+  level, children, size, ...props
+}) =>
+  createElement(LEVEL_TAG[level], props, children))(
```
penx: this is old code but looking at it now I guess it stops us using the `as` prop with `Heading`.

Should we raise an issue for this?

penx:
approved other than 2 comments

components/heading/src/index.js
stevesims: just tried it out as i thought it wouldn't affect things...

When using `<H1 as="h2">` you lose all styles - not good

yes - we should raise a separate issue

Issue #569, opened 2019-02-26
Add proper deprecation support


stevesims:
When attempting to tackle something like #454 it would be good to have instances of the `Header` component complain when they're used, advising that code should be changed to use `Heading` instead.

For reference, this is an example method of deprecating a component via an HOC:
https://whawker.github.io/2016/07/15/deprecating-react-components-with-higher-order-components.html

Additionally we now have some props that are deprecated, such as the old boolean width setting props for `GridCol` - it would be good to have a mechanism for marking props as being deprecated.

For reference, this library provides a mechanism for deprecating props, however it may not be entirely suitable for our current needs:
https://www.npmjs.com/package/react-deprecate

As per #442 this should be added to our documentation.

stevesims:
The `deprecate` HOC now provides a mechanism for marking whole components as deprecated.

At present there is no generic "deprecate a prop" mechanism - instead that is being done on an individual case-by-case basis.

What remains to be done is to document the `deprecate` HOC in `CONTRIBUTING.md`

penx:
duplicate of #442

PR #572, opened 2019-02-26
Feature/deprecation

stevesims:
Adds a `deprecated` HOC

Marks `Header` and `SupportingHeader` components, plus `withWhiteSpace` and `asAnchor` HOCs as being deprecated.

`GridCol` now has custom code to indicate that `columnOneQuarter` style props are deprecated.

Addresses component part of #569 

Builds on feature/heading branch, so currently this is a draft - once that feature has been merged we'll change the PR to go against master.

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged


stevesims:
Custom code added to warn of use of deprecated props in `GridCol` component.

Pondering similar custom code for use of `body` prop in `Table` component

packages/hoc/src/deprecated/index.js
```diff
@@ -0,0 +1,21 @@
+// Deprecation HOC care of:
+// https://whawker.github.io/2016/07/15/deprecating-react-components-with-higher-order-components.html
+import React, { Component } from 'react';
+
+const deprecated = (InnerComponent, msg = '') => class extends Component {
+  componentWillMount() {
+    if (process.env.NODE_ENV === 'development') {
```
penx: we are distributing this code on npm, we process with babel but not webpack before distributing. Can we be sure that the peer application that is running this code will always have a `process` global? And whether it would contain a key `env`?

I guess React does this somehow for their__DEV__ flags but I've not completely figured it out.

https://github.com/facebook/react/issues/770

Have you already looked in to/considered this? Can we find/link to an established pattern for including dev flags when distributing on npm?

One option, though verbose:

`typeof(process) != 'undefined' && process && process.env && process.env.NODE_ENV === 'development'`

stevesims: A good point well raised.

I had thought about this in passing, but essentially only as far as "Will `process` be there? Isn't that a build-time thing? Oh well it works for me in storybook".

I had considered removing the check completely, and *always* producing a warning in the console, irrespective of the build type.  I'm inclined to think that may be the way to go here.

penx: just had a look in a local `node_modules/react/index,js`, and it's this:

https://github.com/facebook/react/blob/b87aabdfe1b7461e7331abb3601d9e6bb27544bc/packages/react/npm/index.js

so thinking if React is doing it then we can do it too.

However I'd match their `=== 'production'` rather than `=== 'development'`

stevesims: i guess in this case it'd be `!== 'production'`

penx: ⭐️

penx:

Building with commit 3e9daa4546342e141c950c55784a0084f82af19c

https://app.netlify.com/sites/govuk-react/deploys/5c76765826bc1e000af40053

PR #575, opened 2019-02-27
Change to docs generation

dsm23:
Issues related to the api-docs/src/index.js were hardcoded to handle unix-like systems. The top-level package.json docs script has been changed to use the command doc-components rather than use it's relative path to the file

**Checklist**:

* [ ] Documentation
* [ ] Tests
* [x] Ready to be merged

### DIFF
--- a/packages/api-docs/src/index.js
+++ b/packages/api-docs/src/index.js
@@ -15,7 +15,7 @@ import generateMarkdown from './markdown/generateMarkdown';
 const components = require('govuk-react');
 
 function getComponentFolderName(file) {
-  const dirs = path.dirname(file).split(path.sep);
+  const dirs = path.dirname(file).split('/');
   let dir = dirs[dirs.length - 1];
   if (dir === 'src' || dir === 'lib') {
     dir = dirs[dirs.length - 2];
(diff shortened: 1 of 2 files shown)


penx:

Building with commit aa44cd762298f3dd2d32ec1c150ad8ab2507ca8c

https://app.netlify.com/sites/govuk-react/deploys/5c76c7947946d0000bf60b02

penx:
Damn me

https://github.com/govuk-react/govuk-react/pull/147/commits/7573cd73e1db1127b45f9c73ebf80c0210771583

For not writing a comment as to why I made this change.

Could have sworn there's a good reason for it. Gonna blame babel 7 beta and that it's something that's now fixed

packages/api-docs/src/index.js
```diff
@@ -74,7 +74,7 @@ async function generateApiForFiles(files) {
 }
 
 export default async function (relDir, outputMd) {
-  const files = await glob(path.resolve(process.cwd(), relDir));
+  const files = await glob(path.resolve(process.cwd(), relDir.replace(/^'/, '').replace(/'$/, '')));
```
penx: this looks a bit mad so could do with a comment, or even better breaking in to a named function with a name that explains its purpose

stevesims: i guess the function could be called `dequote` - and can be simplified a little to be this:
```
function dequote(string) {
  return string.replace(/^'(.*)'$/, '$1');
}
```

PR #583, opened 2019-03-01
Rename Breadcrumb to Breadcrumbs

### LINKED ISSUE #579: Breadcrumb should be called Breadcrumbs
The `Breadcrumb` component is incorrectly named as it shows multiple items, and should therefore be named `Breadcrumbs`.  Existing component should be deprecated.

This would also match govuk-frontend.

stevesims:
This keeps old Breadcrumb for now and just deprecates it

Closes #579

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged

### DIFF
--- a/API.md
+++ b/API.md
@@ -45,32 +45,53 @@ Breadcrumb
 
 ### Usage
 
+This component is DEPRECATED.
+
+Please use the `Breadcrumbs` component instead.
+
+### Properties
+Prop | Required | Default | Type | Description
+:--- | :------- | :------ | :--- | :----------
+ `children` | true | `````` | node | Breadcrumb contents
+
+
+Breadcrumbs
+===========
+
+### Import
+```js
+  import Breadcrumbs from '@govuk-react/breadcrumbs';
+```
+<!-- STORY -->
+
+### Usage
+
 Simple
 ```jsx
-<Breadcrumb>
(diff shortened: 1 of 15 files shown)


penx:

Building with commit 32a2656e9cb4bb35a88287e22286769e705ff0b9

https://app.netlify.com/sites/govuk-react/deploys/5c79601eba71fb0009482cbf

components/breadcrumb/src/index.js
```diff
@@ -106,13 +90,17 @@ const Breadcrumb = ({ children, ...props }) => (
...
   children: PropTypes.node.isRequired,
 };
 
+const Breadcrumb = deprecate(BreadcrumbComponent, 'please use the Breadcumbs component instead');
+
 Breadcrumb.Link = Link;
 
+export { BreadcrumbComponent };
+
 export default Breadcrumb;
```
penx: if this is just a rename, can't we import the BreadCrumbs and wrap it with deprecate?

stevesims: Not quite.

I followed the path of least resistance on this - which has indeed left us with more code than we really need.

The main problem with going for a simpler option of importing the new `Breadcrumbs`, deprecating it, and exporting that would be that it wouldn't be documented.

PR #584, opened 2019-03-01
Add Tabs Component

dsm23:
Addition of Tabs such that they are stylistically similar and can function the same as: [govuk-frontend](https://design-system.service.gov.uk/components/tabs/)

Styled

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged


penx:

Building with commit 7a89bf5fcddb7a2c66029631e4286993942d9fbb

https://app.netlify.com/sites/govuk-react/deploys/5c79728463e04300084bf925

components/tabs/src/atoms/TabList/index.js
```diff
@@ -0,0 +1,26 @@
+import styled from 'styled-components';
```
penx: 🚲 please follow naming convention for files and folders used elsewhere , i.e. `TabList` should be `tab-list`

penx: Related issue raised #587

components/tabs/src/atoms/Tab/index.jsx
```diff
@@ -0,0 +1,101 @@
...
+const StyledHyperLink = styled('a')(
+  typography.font({ size: 19 }),
+  link.common(),
+  link.styleDefault,
+  {
+    display: 'inline-block',
+    paddingTop: spacingSimple2,
+    paddingBottom: spacingSimple2,
+  },
+  ({ isActive }) => ({
```
penx: where a govuk-frontend BEM class is `govuk-tabs__tab--selected` we should aim to:

- have a component 'tabs' ✅
- component 'tabs' should have an atom 'tab' ✅
- atom tab should have prop 'selected' ❌

penx:
- my comments are mainly bikeshedding but please address
- build is failing, so I can't review in Chromatic
- otherwise all good

penx:
an example either manually with context api or an existing and commonly used open source tabs util would be useful but doesn't need to hold up merge

penx:
Another story I'd like to see:

3 tabs, each only loads the context (async) once the tab has been selected. Tab 1 would load by default when the component mounts. However if you're on the mobile breakpoint, all tabs would be shown so all async calls would trigger. The story would need to be built in a way that shows a technique that is compatible with universal rendering.

Again, doesn't need to hold up merge, but a ticket should be created for adding these stories later.

stevesims:
The styling right now when "narrow" (and the tabs disable) is not quite right.

This is a snapshot from govuk-frontend's example:
screen shot 2019-03-07 at 11 56 45

In their version "Contents" is not in bold, and there's a larger gap after the final tab title (Past year)

components/tabs/src/index.js
```diff
@@ -0,0 +1,164 @@
...
+    marginBottom: spacing.responsive({ size: 6, property: 'margin', direction: 'bottom' }),
+  },
+  spacing.withWhiteSpace(),
+);
+
+/**
+ *
+ * ### Import
+ * ```js
+ * import Tabs, { Tab, List, Panel, Title } from '@govuk-react/tabs';
```
stevesims: does this style of import actually work, given how `Tab`, `List`, `Panel` and `Title` are being exported

Also as we have a `Panel` component already it's arguably advisable to just have the example use `Tab.Panel` etc

components/tabs/src/index.js
```diff
@@ -0,0 +1,164 @@
...
+ *         </List>
+ *         <Panel selected={tabIndex === 0} id="first-panel">Panel content 1</Panel>
+ *         <Panel selected={tabIndex === 1} id="second-panel" >Panel content 2</Panel>
+ *       </Tabs>
+ *     );
+ *   }
+ * }
+ * ```
+ *
+ * ##### Hooks Example
```
stevesims: :+1:

would be nice to have this example in the storybook

components/tabs/src/index.js
```diff
@@ -0,0 +1,164 @@
...
+ *  function handleClick({ event: e, index }) {
+ *    const mql = window.matchMedia(`(min-width: ${BREAKPOINTS.TABLET})`);
+ *    if (mql.matches) {
+ *      e.preventDefault();
+ *    }
+ *    return handleTabChange(index);
+ *  }
+ *
+ *  return (
+ * <Tabs>
```
stevesims: indentation here goes a bit awry - but otherwise it's a great example

API.md
```diff
@@ -1875,6 +1875,129 @@ Prop | Required | Default | Type | Description
...
+  // });
+
+  render() {
+    const { tabIndex } = this.state;
+    return (
+      <Tabs>
+        <TabList>
+          <Tab
+            onClick={() => this.setTabIndex(0)}
+            isActive={tabIndex === 0}
```
stevesims: heh - the docs need re-generating 😀

components/tabs/src/index.js
```diff
@@ -0,0 +1,164 @@
...
+Tabs.propTypes = {
+  children: PropTypes.node.isRequired,
+};
+
+Tabs.Tab = Tab;
+Tabs.List = List;
+Tabs.Panel = Panel;
+Tabs.Title = Title;
+
+export { DocumentedTabs };
```
penx: as per @stevesims 's comment above, perhaps we also want to export Tab, List, Panel, Title here (as well as them being appended to component).

I'm not 100% but exporting in both ways may help people who are wrapping Tabs with an HOC and therefore lose Tabs.Tab (not a use case we've come across yet so hard to be sure)

penx:
Approved other than mine and Steve's comments on 
https://github.com/govuk-react/govuk-react/pull/584/files?file-filters%5B%5D=.js#diff-9cae29f918a056445a555b7876c4b564

Also yarn lock needs updating due to conflict

components/tabs/src/index.js
```diff
@@ -0,0 +1,162 @@
...
+Tabs.propTypes = {
+  /** child components that will appear in the Tabs section */
+  children: PropTypes.node.isRequired,
+};
+
+Tabs.Tab = Tab;
+Tabs.List = List;
+Tabs.Panel = Panel;
+Tabs.Title = Title;
+
```
stevesims: exports that were here have gone missing 😢

components/tabs/src/index.js
stevesims: This style of import would not have worked, given the code at the time, as the sub-components were not named exports.

The sub-components are now named exports, but this part of the docs has been removed.

PR #588, opened 2019-03-06
Add VisuallyHidden component and lib

stevesims:
adds new `VisuallyHidden` component, which styles a span as visually-hidden replicating the most common use of that class within govuk-frontend

adds `visuallyHidden()` and `visuallyHidden.focusable()` to `@govuk-react/lib`, tracking equivalent sass helper in govuk-frontend

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged

### DIFF
--- a/API.md
+++ b/API.md
@@ -2092,6 +2092,37 @@ Prop | Required | Default | Type | Description
  `listStyleType` |  | ```undefined``` | string | CSS value for `list-style-type`, or `bullet` or `number` to match govuk-frontend
 
 
+VisuallyHidden
+==============
+
+### Import
+```js
+  import VisuallyHidden from '@govuk-react/visually-hidden';
+```
+<!-- STORY -->
+
+### Usage
+
+This component is primarily intended to be used for material that will be visually hidden
+but visible to screen-reader devices.
+
+Simple
+```jsx
+<VisuallyHidden>Example</VisuallyHidden>
+```
+
+### References
+- https://github.com/alphagov/govuk-frontend/blob/master/src/helpers/_visually-hidden.scss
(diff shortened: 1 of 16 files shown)


penx:

Building with commit 737f4e49f6ba2756d8b28bfa0fb3963770e7e288

https://app.netlify.com/sites/govuk-react/deploys/5c7fffe641498200070fddec

stevesims:
This PR raises an interesting question/issue for Chromatic testing: as this component is only visible when the item is focussed/active is it possible to get Chromatic to focus the element before making it's images?

penx:
> is it possible to get Chromatic to focus the element before making it's images?

yes, we can make a story called 'focussed' that renders the React component and, on did moun  (or similar) of a storybook-only React component wrapper, add client side js in the storybook  to focus the element, then if necessary use the `delay` config prop of Chromatic to wait for it to execute:

http://docs.chromaticqa.com/delay

penx:
as discussed earlier, I'd like to see a story that does e.g.
```jsx
<VisuallyComplete as={Button} />
```

stevesims:
Added in new storybook examples as requested

stevesims:
Hmmm - delayed captures in Chromatic do not appear to have worked, so "focused" versions aren't getting grabbed.  Will make some tweaks

stevesims:
Chromatic still isn't managing to capture these stories where the element should be focussed.  Next step would be to attempt to upgrade to Storybook 5 - however we're on the final Storybook 4 release and the documentation implies that the `delay` config should be working, so it may be a different issue.

For now I'm inclined to leave this issue and attempt to engage with the Storybook support folks to see what they have to say.

PR #592, opened 2019-03-08
Add skip-link component

stevesims:
Replicates skip-link from govuk-frontend.

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged

### DIFF
--- a/components/skip-link/src/stories.js
+++ b/components/skip-link/src/stories.js
@@ -40,7 +40,19 @@ stories.add(
         The skip link component in this example should automatically focus.
         If it loses focus, tab to this example to see it.
       </Paragraph>
-      <AutoFocussed href="#content">Skip to main content</AutoFocussed>
+      <AutoFocussed href="#main-content">Skip to main content</AutoFocussed>
+      <br />
+      <br />
+      <div id="main-content">
+        <Paragraph id="main-content">
+          This paragraph is within a div with an id of `main-content` to match
+          the `href=&quot;#main-content&quot;` prop set on the SkipLink component in this example.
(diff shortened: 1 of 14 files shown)


penx:

Building with commit b4f5bc32c6e74a871ff9feaf822419ce483ed6a6

https://app.netlify.com/sites/govuk-react/deploys/5c8258b9515b0c0009171ccf

components/skip-link/src/stories.js
```diff
@@ -0,0 +1,47 @@
...
+
+stories.add(
+  'Component default',
+  () => (
+    <div>
+      <Paragraph>
+        The skip link component in this example should automatically focus.
+        If it loses focus, tab to this example to see it.
+      </Paragraph>
+      <AutoFocussed href="#content">Skip to main content</AutoFocussed>
```
penx: I think it would be useful to have an example where we're actually linking through to DOM content that exists - i.e. a paragraph with an id below, maybe even showing how SkipLink sits within a standard page structure/markup/component tree.

Also you're providing href="#content" but #content is already the default prop so maybe either provide a custom prop or show it working with the default one.

penx: approved other than this

stevesims: cool - yes - will add another paragraph with an appropriate ID

(FYI - a later PR to this will be to add in an instance of `SkipLink` to the `Page` component - as GDS guidance is that it should be present on every page - added that as #600)

stevesims: docs/examples updated

PR #594, opened 2019-03-11
Fix/windows compat

dsm23:
**Checklist**:

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged

### DIFF
--- a/packages/api-docs/src/index.js
+++ b/packages/api-docs/src/index.js
@@ -73,8 +73,17 @@ async function generateApiForFiles(files) {
   return md;
 }
 
+// unix-like Operating systems requires quotation marks in it's string in the command line
+// whereas windows requires no quotation marks. This approach is to support more kernals
+function dequote(string) {
+  return string.replace(/^'(.*)'$/, '$1');
+}
+
 export default async function (relDir, outputMd) {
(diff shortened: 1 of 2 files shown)


penx:

Building with commit c506941dfd47f3131ae4be944b2351c47186fc3b

https://app.netlify.com/sites/govuk-react/deploys/5c86720e448cbb0007c76cec

packages/api-docs/src/index.js
```diff
@@ -73,8 +73,17 @@ async function generateApiForFiles(files) {
   return md;
 }
 
+// unix-like Operating systems requires quotation marks in it's string in the command line
```
stevesims: not quite true :grin:

the reality is that on a unix shell such as bash, placing quotation marks around strings on the command line stop their values from getting expanded - without these quotes on a bash shell including `**` in a path expression will expand to the first item it finds that's matching.

our command in the `package.json` wraps a path expression in `'` characters to ensure the `**` doesn't get expanded, thus allowing it to still be present in the `relDir` argument below.

bash does not pass on the `'` characters surrounding the command argument, as it recognises them as part of the argument - they get stripped away automatically before the command argument 

windows shells (i think both command.com or powershell) don't expand the `**`, so there the `'` characters aren't necessary.  additionally they don't understand wrapping arguments in `'` characters, so they pass those characters through

PR #599, opened 2019-03-11
Update Jest to v24

dsm23:
The version of jest has been updated to jest 24, jest config has been split out from package.json to its own file. Jest coverageThreshold has been added, this will fail test scripts if coverage falls below a 95% threshold. setupTestFrameowrkScriptFile has been changed to setupFilesAfterEnv which demands an array value. testMatch's relative path has been removed (to better handle micromatch v3, part of jest v24).

Changes to collectCoverageFrom. An extra item has been added to cope with tests in hoc that were not previously being collected. Changes to which filenames can be collected from to add support for would be typescript and jsx. A noteworthy change in jest v24 is that index files whose only purpose was to export object declared as defaults in other files are now being recorded as having zero coverage and this has had asmall impact on overall coverage.

**Checklist**:

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged


penx:

Building with commit 24161994458aa43ee3ec0cc7f74c571288c67c13

https://app.netlify.com/sites/govuk-react/deploys/5c86a225fb0f5a000b991865

package.json
```diff
@@ -112,14 +78,14 @@
...
+    "jest": "^24.4.0",
     "jest-styled-components": "^6.3.1",
     "lerna": "2.11",
     "mkdirp-promise": "^5.0.1",
     "prop-types": "^15.6.0",
-    "react": "^16.2.0",
+    "react": "^16.8.4",
     "react-docgen": "^3.0.0",
-    "react-dom": "^16.2.0",
+    "react-dom": "^16.8.4",
```
penx: I guess this is showing up as a change because you need to rebase off master?

jest.config.js
```diff
@@ -0,0 +1,43 @@
...
+    global: {
+      branches: 95,
+      functions: 95,
+      lines: 95,
+      statements: 95,
+    },
+  },
+  setupFilesAfterEnv: ['<rootDir>/scripts/setupTests.js'],
+  testMatch: [
+    '**/{components,packages}/*/src/**/?(*.)(spec|test).js?(x)',
```
penx: is this the only line that's changed in the config? Would be nice if your PR didn't move the config and change the config at the same time :-)

penx: approved if this line is the only config change :-)

dsm23: Lines 6, 12 - 19 have also been added

package.json
dsm23: I did a  `git merge --no-ff origin/master` then made another commit

jest.config.js
```diff
@@ -0,0 +1,43 @@
+module.exports = {
+  coverageDirectory: './coverage/',
+  collectCoverage: true,
+  collectCoverageFrom: [
+    '<rootDir>/{components,packages}/*/src/**.[jt]s?(x)',
+    '<rootDir>/{components,packages}/*/src/**/*.[jt]s?(x)',
```
stevesims: in principle this line shouldn't be needed - as there's a `**` in the line above that's effectively the same as this one

stevesims: (`**` means "match any character, including `/`")

PR #617, opened 2019-03-15
Allow tabs to be used with react-router

stevesims:
Includes a storybook example

**Checklist**:

* [x] Documentation  (Storybook example added)
* [ ] Tests N/A
* [x] Ready to be merged

### DIFF
--- a/components/tabs/src/atoms/title/index.js
+++ b/components/tabs/src/atoms/title/index.js
@@ -9,4 +9,8 @@ const TabsTitle = styled('h2')(typography.font({ size: 19 }), {
   },
 });
 
+TabsTitle.defaultProps = {
+  children: 'Contents',
+};
+
 export default TabsTitle;
--- a/components/tabs/src/fixtures.js
+++ b/components/tabs/src/fixtures.js
@@ -1,9 +1,11 @@
-import React, { Component } from 'react';
+import React, { Component, useLayoutEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { BREAKPOINTS } from '@govuk-react/constants';
 import { H2, H4 } from '@govuk-react/heading';
 import SectionBreak from '@govuk-react/section-break';
 import Table from '@govuk-react/table';
+import { MemoryRouter, Route, Link } from 'react-router-dom';
+import { useMedia } from 'react-use-media';
 
 import Tabs from '.';
 
@@ -238,7 +240,7 @@ class ProposedClassPropertiesPlugin extends Component {
     const { tabIndex } = this.state;
     return (
       <Tabs>
-        <Tabs.Title>Contents</Tabs.Title>
+        <Tabs.Title />
         <Tabs.List>
           {arrProposedBabel.map(({ contentListItem, id }, index) => (
             <Tabs.Tab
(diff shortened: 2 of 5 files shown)


penx:

Building with commit c163b92e39d10cfc6ca6be33af6b5a1c5b43dfdc

https://app.netlify.com/sites/govuk-react/deploys/5c8bc958ff7f0f00085b4423

components/tabs/src/fixtures.js
```diff
@@ -340,4 +342,60 @@ HooksExample.defaultProps = sharedDefaultProps;
 
 HooksExample.propTypes = sharedPropTypes;
 
-export { HooksExample, ProposedClassPropertiesPlugin, SimpleTabs, SimpleMapTabs, TableTabs };
+// This example shows all panels when in mobile mode, but the route will still be updated
+// to match the last clicked tab title.
+// To show only the active panel on mobiles, simply remove the `isTablet` code.
+// eslint-disable-next-line react/prop-types
+const RouterTabs = ({ match: { params: { section } } }) => {
+  const isTablet = useMedia(`(min-width: ${BREAKPOINTS.TABLET})`);
```
penx: - React components should be pure functions where possible
- I want govuk-react to be usable with universal and server-only React
- I want our fixtures to be representative of real world use cases, they can be simplified but would prefer the examples work universal or server only unless there's a clear reason why this would be a client side only thing

however

- useMedia cannot be used on the server AFAIK
- useMedia means the component is not a pure function AFAIK

so

- if we have to show a client side only solution can we mark it as such and ensure there is an appropriate alternative for server side
- can we just have an example without useMedia? as per your comment "To show only the active panel on mobiles, simply remove the `isTablet` code"

penx: in general I'd prefer we don't introduce a dev dependency on something I consider to be an anti pattern (react-use-media) due to it conflicting with universal rendering and pure functions

stevesims:
Added some further comments for the storybook react-router example, and added two new examples that would be suitable for server-side rendering

components/tabs/src/fixtures.js
stevesims: As per my other comment below, I've adjusted the comments and name for this example, and added two new examples.

I completely understand where you're coming from.  The principle that govuk-react should support SSR is very important.

This particular example is client-side only.  The component inherently needs to know how wide the screen is, and to adjust when it changes.  It should thus be documented accordingly.

I'm not sure I entirely agree that this component is not a "pure function".  Given the same inputs (props and window size), this component should always render the same output.

The grey area comes in given the fact that when the media query listener created by `useMedia` sees a state change, it will then trigger a state update on the component.  That however is the point of effect hooks - to allow for state changes in pure functions.

penx: > I'm not sure I entirely agree that this component is not a "pure function". Given the same inputs (props and window size), this component should always render the same output.

window size is not an input to the function

components/tabs/src/atoms/title/index.js
```diff
@@ -9,4 +9,8 @@ const TabsTitle = styled('h2')(typography.font({ size: 19 }), {
   },
 });
 
+TabsTitle.defaultProps = {
+  children: 'Contents',
```
penx, line 13: Not sure if this is necessary

penx (approved):

PR #623, opened 2019-03-20
Feature/eslint config sonar update

penx:
update to eslint-config-sonar 0.0.7

(also fix minor linting issue raised by eslint-config-sonar@0.0.7)

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "enzyme-to-json": "^3.3.4",
     "eslint": "^4.15.0",
     "eslint-config-airbnb": "^16.1.0",
-    "eslint-config-sonar": "0.0.6",
+    "eslint-config-sonar": "0.0.7",
     "eslint-plugin-flowtype": "^2.42.0",
     "eslint-plugin-import": "^2.8.0",
     "eslint-plugin-jsx-a11y": "^6.0.3",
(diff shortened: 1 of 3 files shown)


penx:

Building with commit 7720c397be2a1e3933ad55fcab55988876d85ec3

https://app.netlify.com/sites/govuk-react/deploys/5c925a194a55b50008a92f03

package.json
```diff
@@ -70,7 +70,7 @@
     "enzyme-to-json": "^3.3.4",
     "eslint": "^4.15.0",
     "eslint-config-airbnb": "^16.1.0",
-    "eslint-config-sonar": "0.0.6",
+    "eslint-config-sonar": "0.0.7",
```
stevesims: wondering if we should put a hat on this

packages/constants/src/typography.js
```diff
@@ -14,7 +14,7 @@ export const TYPOGRAPHY_SCALE = {
     },
     tablet: {
       fontSize: 80,
-      lineHeight: 80 / 80,
+      lineHeight: 1, // 80 / 80
```
stevesims: ah - yes :grin:

package.json
penx: hat wouldn't make a difference at 0.0.x as every change is a breaking change

stevesims: ah - true.

perhaps that means `eslint-config-sonar` should get it's version updated to `0.1.0`?

PR #628, opened 2019-03-21
WIP: Feature/prettier

penx:
I've been changing my mind about Prettier for a while now but have it working on govuk-frontend-react in a way that seems to work seamlessly so keen to get it in to govuk-react.

Why:

- I generally like airbnb's rule set
- I see the need to separate stylistic rules from potential issues
- Whereas I'd be happy to have stylistic rules that don't alter the function of code autofix on save, I wouldn't want other rules to autofix on save.
- create-react-app uses prettier, as do a lot of other React devs I know
- I had previously given up on eslint-config-airbnb and prettier working alongside each other as I was encountering conflicting rules, but it turned out this was due to me not including `"prettier/react"` in extends


penx:

Building with commit 2eec54a98beb158aeaa4d30f015aab824be05a96

https://app.netlify.com/sites/govuk-react/deploys/5c934bd99f9d540007adbee3

PR #629, opened 2019-03-21
Feature/prettier post eslint 5

penx:
I've been changing my mind about Prettier for a while now but have it working on govuk-frontend-react in a way that seems to work seamlessly so keen to get it in to govuk-react.

Why:

- I generally like airbnb's rule set
- I see the need to separate stylistic rules from potential issues
- Whereas I'd be happy to have stylistic rules that don't alter the function of code autofix on save, I wouldn't want other rules to autofix on save.
- create-react-app uses prettier, as do a lot of other React devs I know
- I had previously given up on eslint-config-airbnb and prettier working alongside each other as I was encountering conflicting rules, but it turned out this was due to me not including `"prettier/react"` in extends


penx:

Building with commit b06122fbe8a58a00a0039b50b39e9278519f1f85

https://app.netlify.com/sites/govuk-react/deploys/5c934c32cc6f0400098627df

components/breadcrumb/src/index.js
```diff
@@ -45,7 +45,8 @@ const BreadcrumbListItem = styled('li')({
     position: 'absolute',
     top: '-1px',
     bottom: '1px',
-    left: ((CHEVRON_ALTITUDE_CALCULATED * -2) + CHEVRON_SIZE + CHEVRON_BORDER_WIDTH),
+    left:
+      CHEVRON_ALTITUDE_CALCULATED * -2 + CHEVRON_SIZE + CHEVRON_BORDER_WIDTH,
```
stevesims: this change i'm not keen on - the removing of the outer brackets is OK, but the ones around the multiplication i feel are useful to make it clearer what the calculation is

i generally prefer brackets in maths to clarify, as to me it makes the maths easier to follow

(had this code been written without brackets in the first place then i wouldn't expect this to change)

components/breadcrumbs/src/index.js
```diff
@@ -43,7 +43,8 @@ const BreadcrumbsListItem = styled('li')({
     position: 'absolute',
     top: '-1px',
     bottom: '1px',
-    left: ((CHEVRON_ALTITUDE_CALCULATED * -2) + CHEVRON_SIZE + CHEVRON_BORDER_WIDTH),
+    left:
+      CHEVRON_ALTITUDE_CALCULATED * -2 + CHEVRON_SIZE + CHEVRON_BORDER_WIDTH,
```
stevesims: again, here, for me, this is less comprehensible than it was before

components/button/src/index.js
```diff
@@ -23,18 +19,20 @@ const HALF_SHADOW = RAW_SHADOW / 2;
...
+      weight: isStart ? 'bold' : undefined,
+    }),
   FOCUSABLE,
 
   ({
     buttonColour = BUTTON_COLOUR,
     buttonHoverColour = darken(0.05, buttonColour),
-    buttonShadowColour = (buttonColour === BUTTON_COLOUR)
-      ? BUTTON_COLOUR_DARKEN_15 : darken(0.15, buttonColour),
+    buttonShadowColour = buttonColour === BUTTON_COLOUR
```
stevesims: still not keen on the `?` being on the next line

to me when i'm reading this line i see `buttonShadowColour` being set to a boolean value

components/error-summary/src/fixtures.js
```diff
@@ -8,7 +8,8 @@ import TextArea from '@govuk-react/text-area';
 import ErrorSummary from '.';
 
 export const heading = 'Message to alert the user to a problem goes here';
-export const description = 'Optional description of the errors and how to correct them';
+export const description =
```
stevesims: this change is because of the prettier default line width of 80 characters

i think i still prefer 120

components/breadcrumb/src/index.js
penx: don't think prettier has a config option for this https://prettier.io/docs/en/options.html

components/error-summary/src/fixtures.js
penx: ok will set prettier to use 120

components/fieldset/src/atoms/legend/index.js
```diff
@@ -22,7 +26,7 @@ const StyledLegend = styled('legend')(
   ({ size }) => {
     const actualSize = Number.isNaN(Number(size)) ? HEADING_SIZES[size] : size;
 
-    if ((size !== undefined) && !actualSize) {
+    if (size !== undefined && !actualSize) {
```
stevesims: this is less readable/comprehensible to me

i like use of brackets to make logic explicit.  relying on knowledge of operator precedence makes this stuff harder work

components/grid-col/src/index.js
```diff
@@ -14,20 +14,19 @@ const colValues = {
...
   // No explicit width means auto, so grow/shrink should be set
-  return Object.assign(
-    {},
-    style,
-    {
-      flexGrow: hasAutoWidth ? 1 : 0,
-      flexShrink: hasAutoWidth ? 1 : 0,
-    },
-  );
+  return Object.assign({}, style, {
```
stevesims: not keen on this either

previous version made it clearer what objects where being merged together
this new version is harder to parse

components/breadcrumb/src/index.js
stevesims: yeah - that sucks

this has been raised as an issue in prettier a few times.  unfortunately it seems that there is no solution for now, and that allowing this may not be possible given how prettier is written

components/fieldset/src/atoms/legend/index.js
stevesims: this too is the same issue as with brackets to make mathematical operations clearer

unfortunately prettier just won't leave this stuff alone 😢

PR #630, opened 2019-03-21
Feature/storybook 5

penx:
Update to storybook 5


penx:

Building with commit d5f80be3ca67ead7383a93dcb8897936077622f2

https://app.netlify.com/sites/govuk-react/deploys/5c93c103ce0779000863be50

penx:
Looking at this change:

https://www.chromaticqa.com/snapshot?appId=5ad61aab7f00090020dfdba8&id=5c93c8de17813f00202cb096

I wonder where the flexbox div is coming from. Is it Storybook 5 or storybook-readme 5? If readme, we really don't need the readme wrapping the examples, in fact I don't think we should have this regardless

penx:
ah in fact that's only on the component default where the readme is showing - so I guess it's coming from the readme wrapper but isn't affecting our examples

packages/storybook-components/src/ReadMeHidePreview.js
```diff
@@ -15,11 +15,11 @@ const PreviewComponent = ({ children }) =>
     </div>
   );
 
-const ReadMeHidePreview = (readme, ...rest) =>
-  withDocs({ PreviewComponent })(navigator.userAgent.match(/Chromatic/) ? '' : readme, ...rest);
+const readMeHidePreview = (readme = '') =>
```
stevesims: we should probably change the filename as well - probably to kebab-case

penx: I opened an issue for this #633

components/form-group/src/stories.js
```diff
@@ -11,9 +11,8 @@ import ReadMe from '../README.md';
 const stories = storiesOf('Form/FormGroup', module);
 stories.addDecorator(withKnobs);
 
-stories.add(
-  'Component default',
-  WithDocsCustom(ReadMe, () => (
+stories.add('Component default', () =>
+  WithDocsCustom(ReadMe)(() => (
```
stevesims: should this not be `withDocsCustom`? :wink:

(to be honest, i was confused by this commit, because i'd only very recently added in this story, using this slightly different way of using `withDocsCustom`, and was sure that i'd used the API correctly.  as it goes, i had done things correctly - the underlying `withDocs` API had just changed since, and our API has changed to match it.)

PR #634, opened 2019-03-22
Reduce error summary heading size to match GDS

petergphillips:
GDS uses 24px for the error summary heading - see https://design-system.service.gov.uk/components/error-summary/.  At present error summary uses `H2` which by 36px instead.  This change reduces to `H3` which matches the GDS design.

* [ ] Documentation - N/A
* [x] Tests - Snapshot changed
* [x] Ready to be merged - Yes

Storybook pre change:
storybook pre change

Storybook post change:
storybook post change

GDS Error Summary page:
GDS error summary

### DIFF
--- a/components/error-summary/src/index.js
+++ b/components/error-summary/src/index.js
@@ -4,7 +4,7 @@ import styled from 'styled-components';
 
 import { TEXT_COLOUR, ERROR_COLOUR, FOCUS_COLOUR } from 'govuk-colours';
 
-import { H2 } from '@govuk-react/heading';
+import { H3 } from '@govuk-react/heading';
 import Paragraph from '@govuk-react/paragraph';
 import UnorderedList from '@govuk-react/unordered-list';
 import Link from '@govuk-react/link';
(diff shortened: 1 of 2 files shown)


penx:

Building with commit 3214e927ccb283c51da7a94e375381c5fc105fce

https://app.netlify.com/sites/govuk-react/deploys/5c94a674198ef100077ff300

components/error-summary/src/index.js
```diff
@@ -115,7 +115,7 @@ const StyledErrorSummary = styled('div')(
  */
 const ErrorSummary = ({ onHandleErrorClick, heading, description, errors, ...props }) => (
   <StyledErrorSummary tabIndex={-1} {...props}>
-    <H2>{heading}</H2>
+    <H3>{heading}</H3>
```
stevesims: i've just checked `govuk-frontend` and they actually use an `h2` element in the DOM here.

to better match their markup, this should be changed to `<H3 as="h2">`

petergphillips:
Good spot, not sure how I missed that since the error summary GDS page uses a h2 too :woman_shrugging: 

So the fix would instead be to change the `h2` style to be the correct size instead.  Therefore closing this PR.

PR #636, opened 2019-03-22
Default error summary heading to match GDS guidelines

petergphillips:
GDS guidelines on https://design-system.service.gov.uk/components/error-summary/ states that the heading must be ‘There is a problem’:
error summary heading

We should therefore default the heading to this value.

I haven't changed the docs for the component, since we do still allow the override.  Was thinking about something like:
>Optional message to alert the user to a problem goes here, defaulting to 'There is a problem'

* [ ] Documentation
* [x] Tests
* [x] Ready to be merged

### DIFF
--- a/components/error-summary/src/index.js
+++ b/components/error-summary/src/index.js
@@ -141,7 +141,7 @@ ErrorSummary.propTypes = {
   /** onClick function to scroll the target element into view */
   onHandleErrorClick: PropTypes.func,
   /** Heading text */
-  heading: PropTypes.string.isRequired,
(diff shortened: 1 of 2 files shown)


penx:

Building with commit 66154da4e461e39a4e1bd32cb4c5ae4f09dd9bd3

https://app.netlify.com/sites/govuk-react/deploys/5c94c1eb0aad600007f868db

stevesims:
Cool stuff.

The docs will get updated to show the default value when they're next re-generated.

PR #666, opened 2019-05-23
Modified DateField to allow input type to be set in props

jlrodney:
**Checklist**:

* [X] Documentation
* [X] Tests
* [X] Ready to be merged


penx:

Building with commit ddb86fb7044a75c0cfd24fab3105d87296a14817

https://app.netlify.com/sites/govuk-react/deploys/5ce712b51fb9530008da5d80

penx:
I'd ideally like to see an accompanying test.

Ideally with react-testing-library else with a snapshot.

Other than that, all good! Thanks for the contribution

penx:
also please rebase of master to resolve the yarn.lock file

jlrodney:
Ok great, I'll add a test. Should I squish my commits?

jsrobertson:
Shouldn't be too much of a problem @jlrodney as they'll get squashed when merged.

components/date-field/src/input.js
```diff
@@ -32,7 +32,7 @@ class Input extends React.Component {
         <StyledInput
           name={name}
           error={error}
-          type="number"
+          type={this.props.type ? this.props.type : 'number'}
```
jsrobertson: `number` is already set in the default props.

penx:
I think the intention here was so you don't need type=number, which I've now fixed in another PR so closing this

PR #772, opened 2021-05-15
downgrade docgen and add ci check for yarn docs

### LINKED ISSUE #646: Breakages caused by greenkeeper updates :(
Unfortunately our recent greenkeeper commit that updated several modules has broken a couple of things.

`yarn docs` now shows an error for every component, and produces an empty `API.md` file.  This does not affect `yarn docs:components`.  The error, shown for every component, is in the form:
```
Skipping component: ...govuk-react/components/phase-banner/lib/index.js Configuration contains string/RegExp pattern, but no filename was passed to Babel
```
Upon investigation, this error is thrown within the babel parser.  The error however is not the result of upgrading babel, but rather upgrading `react-docgen` from `^3.0.0` to `^4.1.0`.

The simple fix is therefore to downgrade `react-docgen`.  We may however wish to investigate how to get the newer version to work.

Another problem is that it updated lerna from `2.11` to `~3.13.1`.  Our release script is currently incompatible with lerna 3 (see #447), owing to how it uses `lerna publish` to increment version numbers (which doesn't work with lerna 3 - `lerna version` should be used instead), and how the script blindly continues irrespective of failures on steps.

penx:
Fixes #646

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,7 @@ jobs:
       - run: yarn
       - run: npm install -g codecov
       - run: yarn build:sync
+      - run: yarn docs
       - run: yarn test
       - run: codecov
       - run: yarn build:dist # the dist build is only used to track bundle size delta
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "prettier": "^1.16.4",
     "prop-types": "^15.6.0",
     "react": "^16.8.4",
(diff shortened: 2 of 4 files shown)


penx:

Building with commit 22ad13a148dffb9f20fb2c1da0027a5d87f1d3f2

https://app.netlify.com/sites/govuk-react/deploys/609f2170afe3950007a68dcb

PR #779, opened 2021-05-15
Remove type=number from date field

### LINKED ISSUE #481: DateField type=number causing significant user issues
Quick bit of feedback from a recent change to the `DateField` component - by making the input fields `type=number`, all browsers with HTML5 support add their own version of little ‘up/down’ arrows.

There are two issues with this - firstly, while Chrome makes the arrows nice and small, Firefox and Safari etc. are less elegant and can actually obscure the numbers that users enter.

Second issue is pretty major for our service - people consistently started clicking on the arrows rather than typing into the fields, which immediately caused a lot of grumbling that “It takes ages to select the correct date, and how am I supposed to get it to 2018 in the year field?!“, which is surprising but true across a wide variety of users.

I have temporarily circumvented the issue with this CSS suggestion: https://www.thatstevensguy.com/programming/disable-arrows-on-number-inputs/

screen shot 2018-11-29 at 10 47 43

screen shot 2018-11-29 at 10 46 16

penx:
Fixes #481

### DIFF
--- a/components/date-field/src/__snapshots__/test.js.snap
+++ b/components/date-field/src/__snapshots__/test.js.snap
@@ -538,7 +538,6 @@ exports[`DateField matches wrappersnapshot: wrapper mount 1`] = `
                                 onBlur={[Function]}
                                 onChange={[Function]}
                                 onFocus={[Function]}
-                                type="number"
                               >
                                 <StyledComponent
                                   error={true}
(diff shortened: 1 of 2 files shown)


penx:

Building with commit 9f625a02977f15e4575e9d6e2f4b630f1f1e420c

https://app.netlify.com/sites/govuk-react/deploys/609fee6023dc820008bd7733

PR #784, opened 2021-05-16
add basic storyshots config

stevesims:
non-working right now

adds `coverage:visual-regression` script to `package.json`, as well as `storyshots/index.js` which points to the correct config path

currently fails with an error complaining about an `import` statement.  this most likely indicates a problem with webpack/babel config related to how jest and storybook are interacting


stevesims:
closing in favour of newer version

PR #792, opened 2021-05-16
upgrade to node 16

stevesims:
bump time!


packages/govuk-react/package.json
```diff
@@ -63,7 +63,7 @@
   },
   "peerDependencies": {
     "react": ">=16.2.0",
-    "styled-components": "^4.1.3"
+    "styled-components": ">=5.1"
```
penx, line 66: huh how did I miss that

penx (dismissed):

packages/govuk-react/package.json
stevesims: dunno - I spotted it whilst one of my many attempted builds was happening :grin:

components/header/src/index.js
```diff
@@ -73,4 +73,5 @@ Header.propTypes = {
 export default deprecate(Header, 'please use the Heading component instead');
 
 export { Header as DocumentedHeader };
+// eslint-disable-next-line import/no-cycle
 export { H1, H2, H3, H4, H5, H6 } from './presets';
```
stevesims, line 77: the point of this linting warning is that `presets` is importing this file, thus this is a circular dependency

ideally we'd work around this by having the `Header` component in a separate `atom` file, and import that into the `index` file, allowing the `presets` file to import from the `atom` file, thus ensuring there is no circularity

unfortunately this doesn't work with our documentation generation system.  attempting to do that results in empty documentation

another possibility would be to define H1-H6 inside this file, however that too is incompatible with our documentation generation system, as it results in multiple components being defined inside this file

components/error-summary/src/__snapshots__/test.js.snap
```diff
@@ -93,6 +93,7 @@ exports[`error summary matches the ErrorSummary snapshot: error summary 1`] = `
                 escapeHtml={false}
                 linkRenderer={[Function]}
                 mb={3}
+                parserOptions={Object {}}
```
penx, line 96: odd, any idea what this is?

components/file-upload/src/index.js
```diff
@@ -93,10 +93,13 @@ FileUpload.propTypes = {
     active: PropTypes.bool,
     dirty: PropTypes.bool,
     dirtySinceLastSubmit: PropTypes.bool,
+    // eslint-disable-next-line react/forbid-prop-types
```
penx, line 96: these should be PropTypes.node I think - please could you open an issue to review these?

components/header/src/index.js
penx: please can you open an issue to review this in 0.9?

components/tabs/src/fixtures.js
```diff
@@ -226,16 +238,16 @@ const arrProposedBabel = [
...
-
-  static propTypes = sharedPropTypes;
-
   state = { tabIndex: this.props.defaultIndex };
 
   setTabIndex = setTabIndex;
 
   handleClick = handleClick;
 
+  static propTypes = sharedPropTypes;
```
penx, line 247: why move these?

penx:
approved other than comments

components/error-summary/src/__snapshots__/test.js.snap
stevesims: nope - snapshots be crazy bro!

(it's probably the result of a minor bump in *one* of the modules updated in this PR)

components/tabs/src/fixtures.js
stevesims: linting was saying that `propTypes` had to go after `handleClick`, and the `defaultProps` after `propTypes`.

we seem to have ended up with slightly stricter linting rules with the updated yarn.lock file

components/file-upload/src/index.js
stevesims: added #794

components/error-summary/src/__snapshots__/test.js.snap
penx: maybe react markdown got an upgrade and this is a default prop

penx (approved):

PR #806, opened 2021-05-16
Fix GitHub release

penx:
https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable


stevesims (approved):
cool stuff - I hadn't seen this newer version of setting environment variables inside GitHub Actions - learnt something :grin:

PR #856, opened 2021-05-26
Storybook refactor

### LINKED ISSUE #589: Re-arrange storybook stories
We currently have organised our storybook stories to have "Examples" folders underneath components.  This is done by having a single `storiesOf` declaration for the component itself, and then a separate `storiesOf` declaration for the examples.

Unfortunately as Chromatic just takes the leaf name in `storiesOf` declarations, this approach results in lots of "Examples" in Chromatic that are not tied to their corresponding component.

We should instead add examples to the `storiesOf` declaration for the component.

To facilitate this we'd need to change how we attach documentation to components.  In our `stories.js` files we'd need to remove this line:
```jsx
stories.addDecorator(WithDocsCustom(ReadMe));
```
and instead on our primary/default story do something like this:
```jsx
stories.add(
  'Component default',
  WithDocsCustom(
    ReadMe,
    () => (
      <BackLink goBack={action('go-back')}>{text('Children', 'Back')}</BackLink>
    ),
  ),
);
```

penx:
Fixes #589

## Options considered for category names:

- Core/General/Controls/Common/Basics/UI Elements/Foundation/Components
- Navigation(s)
- Typography/Type
- Content/Data/Communication
- Images/Media/Visualisations
- Form(s)/Inputs
- Layout (e.g. white space, section breaks) Layouts/Screens/Containers/Spacing/Colour
- Icons
- Labs/Experimental
- Utilities/Other
- Deprecated

## Categories

- Forms
- Navigation
- Typography
- Layout
- Utilities


marksy:
Will check in a bit. This pr is massive

marksy:
i would have thought the following would go in navigation:
* Back Link
* Link - navigation, typography

marksy:
i think this structure is best:
Core
Forms & Inputs
Navigation
Typography
Layout
Utilities

marksy:
@penx is there an upgrade path to use Controls over knobs?

marksy:
Would you consider surfacing the constants in Storybook? I added some on my last project which was helpful for reference and click-to-copy the values
image

penx:
I find the problem with having 'Core', if you have a lot of other categories, is it creates uncertainty about where a component can be found.

I've tried to categories everything and remove Core so we now have:

- Form
- Typography
- Page & Layout
- Navigation
- Utility

PR #879, opened 2021-05-31
Use publishConfig instead of workspace-alias

### LINKED ISSUE #877: Use publishConfig pointing to es instead of workspace-alias

penx:
Fixes #877


penx:
Closing for now as this may not be the best approach

PR #902, opened 2021-07-12
add storyshots config

stevesims:
adds a `coverage:visual-regression` script to `package.json`, and a `storyshots/index.js` file to ensure storyshots run

also includes slight modifications to the jest config to ensure the stories work better with jest, and a custom jest config to ensure that coverage thresholds can be loosened a bit

support for `window.matchMedia` for jsdom added to `scripts/setupTests` to allow storyshots to work properly

**Checklist**:

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged


components/footer/src/index.tsx
```diff
@@ -118,7 +118,7 @@ const FooterContainer = styled('footer')(
  * - https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/footer/_footer.scss
  *
  */
-const Footer = ({ children, container: Container, copyright, meta, ...props }) => (
+const Footer = ({ children, container: Container = WidthContainer, copyright, meta, ...props }) => (
```
stevesims, line 121: weird one this - in principle as there's a `defaultProps` value set for `container` this shouldn't be needed, but the story shot default story fails to work properly without this

components/skip-link/src/stories.tsx
```diff
@@ -16,7 +16,7 @@ class AutoFocussed extends Component {
   }
 
   componentDidMount() {
-    this.focusableRef.current.focus();
+    this.focusableRef.current?.focus();
```
stevesims, line 19: in storyshots, the previous version of this fails complaining that `current` is null - this fixes

components/tabs/src/test.tsx
```diff
@@ -3,13 +3,6 @@ import { cleanup, fireEvent, render } from '@testing-library/react';
 
 import { SimpleTabs } from './fixtures';
 
-beforeAll(() => {
```
stevesims, line 6: no longer needed - now inside `setupTests.js`

jest.config.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
   collectCoverageFrom: [
     '<rootDir>/{components,packages}/*/src/**.[jt]s?(x)',
     '<rootDir>/{components,packages}/*/src/**/*.[jt]s?(x)',
-    '!<rootDir>/{components,packages}/*/src/**/{stories,example,fixtures}.[jt]s?(x)',
+    '!<rootDir>/{components,packages}/*/src/**/{stories,example,fixtures,test}.[jt]s?(x)',
```
stevesims, line 11: oddly storyshots was collecting coverage for test files without this 🤷

jest.config.js
```diff
@@ -30,7 +30,7 @@ module.exports = {
   testMatch: ['**/{components,packages}/*/src/**/?(*.)test.[jt]s?(x)'],
   transform: {
     '^.+\\.[jt]sx?$': 'babel-jest',
-    '.+\\.(css|styl|less|sass|scss|png|jpg|ttf|woff|woff2)$': 'jest-transform-stub',
+    '.+\\.(css|styl|less|sass|scss|png|jpg|ttf|woff|woff2|mdx)$': 'jest-transform-stub',
```
stevesims, line 33: gets rid of storyshots complaining about imports of `mdx` files

jest.storyshots.config.js
```diff
@@ -0,0 +1,13 @@
+const config = require('./jest.config');
+
+module.exports = {
+  ...config,
+  coverageThreshold: {
+    global: {
+      branches: 80,
```
stevesims, line 7: slightly lower thresholds here - but these are less important for storyshots

package.json
```diff
@@ -1,30 +1,31 @@
 {
   "name": "@govuk-react/monorepo",
   "scripts": {
-    "start": "cd packages/storybook && npm start",
     "bootstrap": "lerna bootstrap",
-    "docs": "yarn docs:components && concat ./components/*/README.md -o './API.md'",
+    "build:dist": "yarn build:es && rimraf dist && webpack -p packages/govuk-react/es/index.js --devtool source-map --output-filename index.js",
```
stevesims, line 5: sorted these to make them slightly easier to find 😀

package.json
```diff
@@ -1,30 +1,31 @@
...
+    "build:es": "lerna run build:es --parallel",
+    "build:lib": "lerna run build:lib --parallel",
+    "build:storybook": "cd packages/storybook && yarn build",
+    "build:sync": "lerna run build",
+    "build": "lerna run build --parallel",
+    "bundlesize": "bundlesize",
+    "clean:workspace": "rimraf packages/*/lib && rimraf packages/*/es && rimraf components/*/lib && rimraf components/*/es",
+    "clean": "rimraf lib && rimraf es",
+    "coverage:visual-regression": "jest --watchAll=false --coverage --coverageDirectory=coverage/visual-regression --roots=\"<rootDir>\" --testMatch=\"<rootDir>/storyshots/index.js\" --config=\"./jest.storyshots.config.js\"",
```
stevesims, line 14: this is the only additional command

penx: annoyingly, only stories that are enabled in chromatic actually constitute coverage...

do you think we can/should make this calculate accordingly?

side note, I should reenable chromatic...

components/footer/src/index.tsx
penx: this should be looked in to, the only thing I can think of right now is that the prop is

`container: Footer.WidthContainer`

perhaps make this `container: WidthContainer` so it still works if reordered?

penx: I think I'm happy with this PR as is other than this issue, and I want to run it through locally.

stevesims: it's odd - the `Page` container uses the same pattern, and the `defaultProps` value for `container` seems to work fine there

penx: this appears to be a bug with Storybook 6.3, there is a conversation on this here:

https://github.com/storybookjs/storybook/issues/15378#issuecomment-878968025

package.json
```diff
@@ -47,6 +48,7 @@
     "@storybook/addon-actions": "6.2.9",
     "@storybook/addon-docs": "6.2.9",
     "@storybook/addon-knobs": "6.2.9",
+    "@storybook/addon-storyshots": "^6.3.4",
```
penx, line 51: storybook/addon-storyshots version should match other storybook version numbers (6.2.9)

components/footer/src/index.tsx
penx: Hopefully rolling back storyshots version to 6.2.9 will resolve this issue and this change won't be needed.

Also worth noting that defaultProps on functional components have been deprecated, so we'll need to deal with that at some point anyway, just I'd prefer not to do it as part of this PR.

penx: rolling back doesn't seem to fix the issue, will keep looking in to it

penx: ok so

- I think this is a bug in storybook 6.3, in that props can be set to undefined even if a default prop exists
- the same bug also exists in earlier versions of storyshots
- however, as default props on functional components are deprecated, I'm fine with us moving away from them in order to work round this bug

please can you remove all defaultProps in Footer and only use es6 defaults?

penx: although this doesn't explain why it's working on the Page component, will look in to that

penx: footer is exporting the component itself as the default story

`export const Default = Footer;`

however Page is wrapping this

```
export const Default = () => (
  <div>
    <Page beforeChildren={<BackLink href="#" />}>
      <H1>Page Title</H1>
    </Page>
  </div>
);
```

which is likely the reason it's working for Page and not Footer

If we change the default Page story to `export const Default = Page;` then we get the same issue.

stevesims: ok cool - I'll remove the `defaultProps` from footer - I wasn't really aware that they were deprecated, but I do tend to prefer using es6 defaults, especially in typescript

medium term we should probably look to remove `propTypes` too and use TS types instead - but that's for another time

penx: Yeah it had snuck me by too https://github.com/facebook/react/pull/16210

penx (approved):
@stevesims approved but there are some linting issues that need fixing

stevesims:
yeah - I did spot the linting issue - my fix to the footer component triggers the rule - doh!

my instant thought was to disable the rule, possibly globally, as I'm not convinced the rule is useful.  haven't managed to give it much thought yet tho - blanket disabling the rule feels like it might be a bit drastic

components/footer/src/index.tsx
stevesims: there's a new note that's been added to the storybook bug conversation showing how this can be fixed by changing the default story for the Footer, so I've now restored the `defaultProps`, removed this es6 default, and changed the story

if/when we move from `propTypes` to TS type declarations then we can put the es6-style value default back in

stevesims:
new comment was added to the storybook thread https://github.com/storybookjs/storybook/issues/15378#issuecomment-884668183, so have updated this accordingly, restoring the `defaultProps` and changing the default story definition to fix the previous error

PR #906, opened 2021-07-23
merge unit and regression coverage

stevesims:
adds a `coverage` command to package.json (and several support commands) that will generate a merged coverage report from both unit tests and visual regression tests (storyshots)

**Checklist**:

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged


package.json
```diff
@@ -123,6 +128,21 @@
...
+    "branches": 95,
+    "functions": 90,
+    "lines": 95,
+    "statements": 95,
+    "watermarks": {
+      "branches": [90, 95],
+      "functions": [90, 90],
+      "lines": [90, 95],
+      "statements": [85, 95]
+    }
```
stevesims, line 144: these values are all kinda arbitrary for now

the purpose of this is to ensure that we get red (and yellow) colouring in the report for individual files

stevesims:
hmm - dependabot automatically updated @storybook/react which has screwed up our storybook - there's no stories being shown any more 😞

stevesims:
updating all the storybook modules seems to fix storybook (seemed not to work for me for a while - there were a few hiding that I didn't spot - doh!)

packages/storybook/package.json
```diff
@@ -49,11 +49,11 @@
...
-    "@storybook/addon-docs": "6.2.9",
-    "@storybook/addon-knobs": "6.2.9",
-    "@storybook/addon-links": "6.3.2",
-    "@storybook/addons": "6.2.9",
+    "@storybook/addon-actions": "6.3.5",
+    "@storybook/addon-docs": "6.3.5",
+    "@storybook/addon-knobs": "6.3.0",
+    "@storybook/addon-links": "6.3.5",
+    "@storybook/addons": "6.3.5",
     "@storybook/react": "6.3.5",
```
penx, line 57: hmm we're on 6.3.5 already are we? 🤔 

I take it you've searched for "6.2.9" and updated everywhere?

Why is addon knobs only updated to 6.3.0?

penx:
> hmm - dependabot automatically updated @storybook/react which has screwed up our storybook - there's no stories being shown any more 😞

Is there something our CI should have done to catch that?

penx:
approved other than

- question on knobs being 6.3.0
- any quick win on getting CI to have prevented the storybook auto merge? Or create a ticket?

packages/storybook/package.json
stevesims: I tried to update addon-knobs too, but found it's been deprecated and 6.3.0 is the latest version - replacement is addon-controls

stevesims:
yeah I think we should probably create a ticket to think a bit more about potentially preventing future issues with rogue storybook (or other) module updates causing problems.

for the storybook update, it passed as chromatic didn't seem to get CI checked for that PR - https://github.com/govuk-react/govuk-react/runs/3141899601 - which makes some sense as otherwise dependabot would thrash our chromatic usage, but conversely if we did check chromatic for dependabot storybook module updates (only) that would have caught this

packages/storybook/package.json
penx: oh right

penx:
> yeah I think we should probably create a ticket to think a bit more about potentially preventing future issues with rogue storybook (or other) module updates causing problems.
> 
> for the storybook update, it passed as chromatic didn't seem to get CI checked for that PR - https://github.com/govuk-react/govuk-react/runs/3141899601 - which makes some sense as otherwise dependabot would thrash our chromatic usage, but conversely if we did check chromatic for dependabot storybook module updates (only) that would have caught this

#907

penx (approved):

Issue #81, opened 2018-02-27
Remove JSX Wrappers from documented styled components


penx:
In https://github.com/penx/govuk-react/pull/79 some components have been renamed from e.g. `Button` to `GButton` (edit, now `Styled*`) and then wrapped with a JSX wrapper so that they are picked up by react-docgen.

We ideally don't want to dictate code structure due to documentation generation so want to look in to another solution in the future.

https://github.com/reactjs/react-docgen/issues/256

Loque-:
Not sure I fully understand this one, has it been resolved? I cannot see this issue in the repo currently.

penx:
Yes this issue is still present in the repository,

e.g.

```jsx
const ErrorText = props => <StyledError {...props} />;
```

This is redundant code. We could just export StyledError but need to wrap it with JSX so that it is picked up by react-docgen.

Doing a find on ` = props => <Styled` throughout the whole project should show all places this is happening.

penx:
Related

https://github.com/reactjs/react-docgen/issues/256
https://github.com/jquense/react-docgen-styled-resolver/issues/1

penx:
react-docgen-styled-resolver would be good but I haven't been able to get it to work

https://github.com/jquense/react-docgen-styled-resolver/issues/1

penx:
We now have docs appearing in Storybook via their docs addon.

https://github.com/storybookjs/storybook/tree/master/addons/docs
https://github.com/storybookjs/storybook/tree/master/addons/docs/src/lib/docgen

In part, they seem to be using

https://github.com/storybookjs/babel-plugin-react-docgen

penx:
Unfortunately storybook's docs generation relies on the same code and has the same issue

https://github.com/storybookjs/storybook/issues/8279

After converting to typescript, Storybook uses react-docgen-typescript, but this doesn't seem to render any docs at all so I have reverted to react-docgen as described at:

https://gist.github.com/shilman/036313ffa3af52ca986b375d90ea46b0

penx:
As we are reliant on Storybook generation, and I don't think we can configure Storybook to use a custom resolver, I don't think this is going to be fixable even if we were to write a custom resolver, until storybookjs/storybook#8279 is resolved.

penx:
This was resolved during the transition to TypeScript

PR #909, opened 2021-07-27
migrate more tests to @testing-library/react

stevesims:
migrates tests for the following components from enzyme to @testing-library/react:
Caption
Checkbox
Details
ErrorSummary
ErrorText
Fieldset
FileUpload
FormGroup
GridCol
GridRow
Heading
HintText

all updated tests no longer check implementation details, just functionality, and have had all snapshots removed

tests for ErrorSummary rewritten, and minor changes made to the component to make it more resiliant and have higher coverage

bug fix to ErrorSummary component - `defaultProps` had been re-defined, which meant that not providing an `errors` array could cause it to crash - the definition has now been fixed

**Checklist**:

* [ ] Documentation N/A
* [x] Tests
* [x] Ready to be merged


components/error-summary/src/index.tsx
```diff
@@ -155,8 +156,4 @@ ErrorSummary.propTypes = {
   ),
 };
 
-ErrorSummary.defaultProps = {
-  heading: 'There is a problem',
-};
-
```
stevesims, line 161: this was overwriting the `defaultProps` set above - oops!

components/error-summary/src/index.tsx
```diff
@@ -123,7 +123,7 @@ const ErrorSummary = ({ onHandleErrorClick, heading, description, errors, ...pro
       <UnorderedList mb={0} listStyleType="none">
         {errors.map((error, index) => (
           <ListItem key={error.targetName}>
-            <StyledErrorText tabIndex={index + 1} onClick={() => onHandleErrorClick(error.targetName)}>
+            <StyledErrorText tabIndex={index + 1} onClick={() => onHandleErrorClick?.(error.targetName)}>
```
stevesims, line 126: seems this is our first use of the optional chaining operator inside a component (we've used it elsewhere inside some stories and examples)

the ancient version of react-docgen we had been using didn't like this, which broke CI - fortunately it looks like the latest release works OK

out of curiosity I tried removing the hack we have for ensuring components get documented to see if that issue is resolved with the latest react-docgen - no joy

penx: some more details on that issue here https://github.com/govuk-react/govuk-react/issues/81#issuecomment-841740534

components/error-summary/src/index.tsx
penx: looks like there's no change to behaviour though? So no need to mark this as 'bugfix' or anything?

components/checkbox/src/test.tsx
```diff
@@ -1,35 +1,29 @@
...
+    render(<Checkbox>Example</Checkbox>);
   });
 
   it('should render an input type="checkbox"', () => {
-    const output = shallow(wrapper);
-    expect(output.find('input[type="checkbox"]')).toBeTruthy();
-  });
+    const { getByRole, getByText } = render(<Checkbox>Example</Checkbox>);
 
-  it('should render a label', () => {
```
penx, line 21: do we not want this test anymore?

penx: maybe a test using getByLabel?

components/error-summary/src/test.tsx
```diff
@@ -1,86 +1,52 @@
...
 import ErrorSummary from '.';
-import ErrorSummaryExample, { heading, description, errors } from './fixtures';
 
 describe('error summary', () => {
-  const wrapperErrorSummary = mount(<ErrorSummaryExample />);
-
-  it('renders without crashing', () => {
-    const div = document.createElement('div');
-    ReactDOM.render(<ErrorSummaryExample />, div);
+  it('should render without crashing', () => {
```
penx, line 7: I think tests like this are giving false code coverage metrics and should be removed or updated to assert key features of a default render

penx: https://github.com/govuk-react/govuk-react/issues/910

components/footer/src/index.tsx
```diff
@@ -169,7 +169,7 @@ Footer.WidthContainer = WidthContainer;
 
 Footer.defaultProps = {
   children: undefined,
-  container: WidthContainer,
+  container: Footer.WidthContainer,
```
penx, line 172: any reason this was changed?

components/form-group/src/test.tsx
```diff
@@ -1,14 +1,20 @@
 import React from 'react';
-import { mount } from 'enzyme';
+import { render } from '@testing-library/react';
 
 import { FormGroupDocumented as FormGroup } from '.';
 
 describe('FormGroup', () => {
-  it('matches snapshot', () => {
-    expect(mount(<FormGroup>FormGroup example</FormGroup>)).toMatchSnapshot('FormGroup');
+  it('renders contents without crashing', () => {
```
penx, line 7: I'm not sure this is what we want by 'convert enzyme to testing library'.

I'd rather take it slower but make sure the tests are written well in the process.

i.e. written from the perspective of the user, not testing implementation details and not giving false code coverage metrics due to vague assertions

components/error-summary/src/index.tsx
stevesims: it was essentially a hidden (non-reported) bug - I think the only affect of it was that if no errors were given the component would crash (as the defaulting of the `errors` prop to `[]` was previously being lost)

components/checkbox/src/test.tsx
stevesims: hmm - I kinda looked on this as essentially being covered by other tests, plus also this being an implementation detail

components/error-summary/src/test.tsx
stevesims: makes sense, and I'd already been thinking roughly along those lines (I just hadn't quite got that far :grin:) - several of these tests are now checks that they essentially pass thru children which, for most of them, is their default behaviour

with that in mind, given the rest of the coverage in this file, this particular test can be deleted

components/footer/src/index.tsx
stevesims: yes - this was a change that got made fairly recently, IIRC attempting to chase down a phantom bug with storyshots.

the previous version meant that the documentation inside storybook said that the default value for this prop was `WidthContainer`, but there's no `WidthContainer` component directly exported.  in reality a user would need to use `Footer.WidthContainer`, so this change (back) means that the documentation is now more correct

components/form-group/src/test.tsx
stevesims: I hear you, but I'm not sure what else we can test for this component

the FormGroup component essentially just renders content, putting it inside a container.  there's support for an `error` prop which adjusts the display.  that's it.

I guess there's an argument that the second test here that checks for a successful render with the `error` flag set could be deleted, and instead we could/should rely on storybook for coverage of that

components/checkbox/src/test.tsx
stevesims: I've changed things around a bit so this is now explicitly covered

components/error-summary/src/index.tsx
penx: ok have added bug and tests labels to the PR so we know

components/checkbox/src/test.tsx
penx: thanks, being a label isn't an implementation detail - to users, it's an accessibility feature, using react testing library I guess this is byLabelText https://testing-library.com/docs/queries/bylabeltext/

components/error-summary/src/test.tsx
penx: agreed please either:
- delete tests like this rather than migrate
- reword them and add assertions to cover effects of default props

components/error-summary/src/index.tsx
penx: minor - I'm thinking that this should be mentioned in the PR description as it's relevant to release notes

components/checkbox/src/test.tsx
stevesims: yup - revised test uses `getByLabelText`

components/error-summary/src/index.tsx
stevesims: updated

components/error-summary/src/test.tsx
stevesims: all the tests that have been edited so far have either had their "should render without crashing" test deleted, replaced with a "renders contents without crashing" that asserts that children passed to the component get rendered, or replaced with other tests for default props.

for now, or CI hasn't been updated to use the merged coverage - I wasn't quite sure how to adjust that.  so far tho coverage has not dropped for unit tests

there are some components with updated tests that have default props where these things are not explicitly tested in unit tests - these are all, I believe, setting a default `size` props to give a default font size.  I'm inclined to think of that as an implementation detail, and best checked via storybook.  an explicit test for "this is size XLARGE by default" feels wrong - and a test of "has a default size" also feels not quite right

penx: > setting a default size props to give a default font size

Yes, I'd put it like this:

Any props (or similar forks) that, as far as the user is concerned, only effects the visual output (such as font size and spacing), should be tested via visual regression and we shouldn't be concerned about coverage via unit tests. The important thing to note is there should be unit tests where there is a significant accessibility functionality such as 'labeled by' as this is not visual-only.

components/breadcrumbs/src/atoms/link/test.tsx
```diff
@@ -4,7 +4,10 @@ import { render } from '@testing-library/react';
 import Link from '.';
 
 describe('Breadcrumbs Link', () => {
-  it('renders without crashing', () => {
-    render(<Link href="#id">Example</Link>);
+  it('renders contents without crashing', () => {
+    const contents = 'Example';
+    const { getByText } = render(<Link href="#id">{contents}</Link>);
+
+    expect(getByText(contents)).toBeInTheDocument();
```
penx, line 11: minor - I would suggest for this test, from the user's perspective, they would expect to be able to click on the contents (even if nested dom contents) and the url would update to include #id

components/back-link/src/test.tsx
```diff
@@ -5,12 +5,20 @@ import userEvent from '@testing-library/user-event';
 import BackLink from '.';
 
 describe('Back Link', () => {
-  it('renders without crashing', () => {
+  it('renders contents without crashing', () => {
```
penx, line 8: same comment as https://github.com/govuk-react/govuk-react/pull/909/files#r680359241

components/breadcrumbs/src/atoms/link/test.tsx
penx: in the process of writing this test from the users perspective, it would also test that it 'renders contents without crashing'

components/fieldset/src/test.tsx
```diff
@@ -1,10 +1,13 @@
 import React from 'react';
-import { mount } from 'enzyme';
+import { render } from '@testing-library/react';
 
 import { FieldsetDocumented as Fieldset } from '.';
 
 describe('Fieldset', () => {
-  it('matches snapshot', () => {
-    expect(mount(<Fieldset />)).toMatchSnapshot('Fieldset');
+  it('renders contents without crashing', () => {
```
penx, line 7: from the perspective of the user, I think this test would read 'contents is visible'

components/hint-text/src/test.tsx
```diff
@@ -1,19 +1,13 @@
...
-import ReactDOM from 'react-dom';
-import { mount } from 'enzyme';
+import { render } from '@testing-library/react';
 
 import HintText from '.';
 
 describe('HintText', () => {
-  const example = 'example';
-  const wrapper = <HintText>{example}</HintText>;
+  it('renders contents without crashing', () => {
```
penx, line 7: same comment as https://github.com/govuk-react/govuk-react/pull/909/files#r680359456

components/heading/src/test.tsx
```diff
@@ -32,54 +31,50 @@ describe('Heading', () => {
     process.env = OLD_ENV;
   });
 
-  it('renders a Heading and all the H-level tags without crashing or producing a warning', () => {
+  describe('renders a Heading and all the H-level tags without crashing or producing a warning', () => {
```
penx, line 34: same comment as https://github.com/govuk-react/govuk-react/pull/909/files#r680359456

components/grid-row/src/test.tsx
```diff
@@ -1,18 +1,13 @@
 import React from 'react';
-import ReactDOM from 'react-dom';
-import { mount } from 'enzyme';
+import { render } from '@testing-library/react';
 
 import GridRow from '.';
 
 describe('GridRow', () => {
-  const wrapper = <GridRow>Example</GridRow>;
+  it('renders contents without crashing', () => {
```
penx, line 7: same comment as https://github.com/govuk-react/govuk-react/pull/909/files#r680359456

components/form-group/src/test.tsx
penx: maybe name the test 'renders contents' and remove 'without crashing'

penx:
approved other than comments

components/breadcrumbs/src/atoms/link/test.tsx
penx: @stevesims any thoughts on this and https://github.com/govuk-react/govuk-react/pull/909#discussion_r680850550 ?

penx:
@stevesims I'm going to merge this as there hasn't been any activity and I'd rather the changes go in as-is rather than be lost over time.

PR #920, opened 2021-08-25
Fix typings on backLink

marksy:
* add types to `font()`
* make arg optional for `common()`
* rename `setupTests.js` to `setupTests.ts`

**Checklist**:
* [ ] Documentation NA
* [ ] Tests NA
* [x] Ready to be merged

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -24,7 +24,7 @@ module.exports = {
     },
   },
   moduleFileExtensions: ['web.js', 'js', 'json', 'web.jsx', 'jsx', 'node', 'ts', 'tsx'],
-  setupFilesAfterEnv: ['<rootDir>/scripts/setupTests.js'],
+  setupFilesAfterEnv: ['<rootDir>/scripts/setupTests.ts'],
   snapshotSerializers: ['enzyme-to-json/serializer'],
   testEnvironment: 'jsdom',
   testMatch: ['**/{components,packages}/*/src/**/?(*.)test.[jt]s?(x)'],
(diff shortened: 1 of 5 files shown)


jest.config.js
```diff
@@ -24,7 +24,7 @@ module.exports = {
     },
   },
   moduleFileExtensions: ['web.js', 'js', 'json', 'web.jsx', 'jsx', 'node', 'ts', 'tsx'],
-  setupFilesAfterEnv: ['<rootDir>/scripts/setupTests.js'],
+  setupFilesAfterEnv: ['<rootDir>/scripts/setupTests.ts'],
```
penx, line 27: I think you forgot to commit setupTests.ts

packages/lib/src/typography/index.tsx
```diff
@@ -53,7 +53,14 @@ export function responsive(size, overrideLineHeight) {
   };
 }
 
-export function font({ size, weight = 'regular', tabular = false, lineHeight } = {}) {
+interface FontProps {
```
penx, line 56: this isn't a React component so I'd prefer FontOptions to FontProps

penx (changes requested):
couple of comments, otherwise good

tsconfig.json
```diff
@@ -8,6 +8,9 @@
     "jsx": "react",
     "esModuleInterop": true
   },
+  "include": [
+    "./scripts/setupTests.ts"
```
penx, line 12: @marksy do you remember why this was needed? It broke other typescript features, e.g. you can no longer command click/go to definition. Removing this fixes the issue.

marksy: For jest set up iirc

PR #938, opened 2021-09-05
Review use of forwardRef

### LINKED ISSUE #921: Ability to pass refs to all form fields
React hook form requires a ref to be passed to the input

penx:
Fixes  #921

> components that take an input prop and spread it to the input, don't need forwardRef as it can be done via the input prop.
>
> I'll remove it for InputField, add it to SearchBox and not touch Select and TextArea

https://github.com/govuk-react/govuk-react/issues/921#issuecomment-913180653

### DIFF
--- a/API.md
+++ b/API.md
@@ -295,7 +295,7 @@ With custom input name props
 Prop | Required | Default | Type | Description
 :--- | :------- | :------ | :--- | :----------
  `children` | true | `````` | node | 
- `defaultValues` |  | ```{   day: undefined,   month: undefined,   year: undefined, }``` | custom | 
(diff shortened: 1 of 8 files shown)


marksy:
You have a couple snapshots needing to be updated

marksy (commented):
Is the api.md generated?

`shape [object object]` should be more explicit about what’s in the object?

penx:
> Is the api.md generated?
> 
> `shape [object object]` should be more explicit about what’s in the object?

Yep it's generated. I feel it's not exactly worse than what it was before, i.e. 'custom'. I'm not too fussed about it and feel it could be a time sink trying to find something that generates neater docs.

The 'default value' next to it gives a good idea of what the shape should be.

PR #1048, opened 2021-11-10
Set storybook to be based on light theme

davidwinter:
If your OS (such as macOS) is set to use the dark theme, some browsers (like Firefox) pass this through to Storybook which will then use the dark theme as default. This means components will then not have a white background when viewing them within Storybook.

This PR sets the base theme to be the `light` Storybook theme, regardless of what your OS setting is, as the components are meant to be placed on a white background.

Changed from:

image

To:

image

**Checklist**:

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged

### DIFF
--- a/packages/storybook/.storybook/manager.js
+++ b/packages/storybook/.storybook/manager.js
@@ -3,6 +3,7 @@ import { addons } from '@storybook/addons';
 import { create } from '@storybook/theming';
 
 const theme = create({
+  base: 'light',
   brandTitle: 'govuk-react',
   sortStoriesByKind: true,
   brandUrl: 'https://github.com/govuk-react/govuk-react',


penx:
Thanks! I think I'd prefer the canvas background to be set to white, and still allow storybook to be dark theme, but this is definitely an improvement as-is so will merge it in.

davidwinter:
@penx possible to have this deployed to GitHub Pages before the next version release?

https://govuk-react.github.io/govuk-react

penx:
@davidwinter that should be done, are you able to test and let me know?

davidwinter:
Sorry for delay @penx - just to confirm this is now working as expected :+1:

PR #1057, opened 2022-01-28
Update GDS panel component colour to the green

Trunk89:
Simple update to the colour of GDS panel component

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged

### DIFF
--- a/components/panel/src/index.tsx
+++ b/components/panel/src/index.tsx
@@ -5,7 +5,7 @@
  */
 import * as React from 'react';
 import styled from 'styled-components';
-import { TURQUOISE, WHITE } from 'govuk-colours';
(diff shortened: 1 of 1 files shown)


components/panel/src/index.tsx
```diff
@@ -31,7 +31,7 @@ const StyledPanel = styled('div')(
     },
 
     color: WHITE,
-    background: TURQUOISE,
+    background: BUTTON_COLOUR,
```
penx, line 34: Sorry I never got back on this one. I don't think we should be using BUTTON_COLOUR here.

If you still need this, please can you reference what the issue in the PR description? Is govuk-react out of sync with the design system here? Perhaps add some links or screenshots?

Can we access the colour we need from govuk-colours by another name?

PR #1075, opened 2022-04-30
Fix Netlify errors by migrating to bundlesize2

domdomegg:
**Checklist**:

* [x] Documentation (N/A)
* [x] Tests (N/A)
* [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "babel-jest": "^27.0.2",
     "babel-plugin-add-module-exports": "^1.0.4",
     "babel-plugin-styled-components": "^1.12.0",
-    "bundlesize": "^0.18.1",
+    "bundlesize2": "^0.0.31",
     "can-npm-publish": "1.3.3",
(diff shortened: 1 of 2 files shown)


domdomegg:
Meant to leave a better comment on this one...

Currently govuk-react uses bundlesize to make sure the bundle size is reasonable so it can be installed and used efficiently by downstream consumers. bundlesize depends on iltorb, which has been deprecated and has had some releases removed from its repo. This breaks the netlfiy build (specifically: when netlify tries to install dependencies, it needs to recompile iltorb from scratch, which requires Python, which it doesn't have - and therefore errors).

This PR replaces bundlesize with bundlesize2, which is largely a drop-in replacement. It's made by the same author, and has been indicated to be the successsor to bundlesize. It doesn't have the same iltorb problems as bundlesize.

Also see https://github.com/siddharthkp/bundlesize/pull/370

penx (approved):

PR #1072, opened 2022-04-30
Allow footer licence customization

domdomegg:
Closes https://github.com/govuk-react/govuk-react/issues/1059

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged

**Screenshots**:

image

image

PR #1073, opened 2022-04-30
Regenerate docs to keep up to date

domdomegg:
**Checklist**:

* [x] Documentation
* [x] Tests (N/A)
* [x] Ready to be merged

I ran `yarn docs` on the repo and these automatically updated. Looks like React's definitions updated which then are reflected in the Paragraph docs.

PR #1076, opened 2022-04-30
Improve doc generation

domdomegg:
**Checklist**:

* [x] Documentation
* [x] Tests (N/A)
* [x] Ready to be merged (see notes below)

**Interaction with other PRs**:

This PR can be merged instead of #841. We might also want to regenerate the docs after #1072, given it uses union types too. Alternatively, if that's merged first then we can easily remerge in master and overwrite the conflicts by rerunning `yarn docs` again.

## Escaping pipes

Currently when we have union types, for example `string | null`, we're generating doc tables that can look like:

```markdown
Prop | Required | Default | Type          | Description
:--- | :------- | :------ | :------------ | :--------------
text | true     |         | string | null | The text body to display
```

Unfortunately this breaks our docs, as it renders as:

Prop | Required | Default | Type          | Description
:--- | :------- | :------ | :------------ | :--------------
text | true     |         | string | null | The text body to display

(e.g. see https://www.npmjs.com/package/@govuk-react/unordered-list `margin`)

This PR changes the logic so the pipe is escaped, so we output:

```markdown
Prop | Required | Default | Type           | Description
:--- | :------- | :------ | :------------- | :--------------
text | true     |         | string \| null | The text body to display
```

Which results in:

Prop | Required | Default | Type           | Description
:--- | :------- | :------ | :------------- | :--------------
text | true     |         | string \| null | The text body to display

## Fixing defaults

Currently when we don't have a default, we're generating doc tables that can look like:

```markdown
Prop | Required | Default | Type          | Description
:--- | :------- | :------ | :------------ | :--------------
text | true     | ``````  | string        | The text body to display
```

Unfortunately this breaks our docs, as it renders as:

Prop | Required | Default | Type          | Description
:--- | :------- | :------ | :------------ | :--------------
text | true     | ``````  | string        | The text body to display

(e.g. see https://www.npmjs.com/package/@govuk-react/unordered-list `children`)

This PR changes the logic so when there is no default, we don't wrap the prop default in backticks, so we output:

```markdown
Prop | Required | Default | Type          | Description
:--- | :------- | :------ | :------------ | :--------------
text | true     |         | string        | The text body to display
```

Which results in:

Prop | Required | Default | Type          | Description
:--- | :------- | :------ | :------------ | :--------------
text | true     |         | string        | The text body to display


penx:
> This PR can be merged instead of https://github.com/govuk-react/govuk-react/pull/841

I think you meant a different PR as #841 was merged a year ago?

domdomegg:
> > This PR can be merged instead of #841
> 
> I think you meant a different PR as #841 was merged a year ago?

Yep, good catch. I meant #1073, sorry.

penx (approved):

PR #1077, opened 2022-04-30
Add support for Tag tints

domdomegg:
**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged

- Algins padding with GDS: https://design-system.service.gov.uk/components/tag/
- Adds tints as per GDS: https://design-system.service.gov.uk/components/tag/#additional-colours (NB: in GDS these are specific colours for tags, and not part of the wider [GDS colour pallete](https://design-system.service.gov.uk/styles/colour/). This is reflected in the code here, where it is used specifically for Tag)

**Screenshots**:

image

image

image


components/tag/src/index.tsx
```diff
@@ -16,34 +34,57 @@ import { typography } from '@govuk-react/lib';
...
+  backgroundColor?: string;
+  /**
+   * Custom colour to override the tag text color
+   */
+  color?: string;
 }
 
 Tag.defaultProps = {
   inactive: false,
+  tint: 'SOLID',
```
penx, line 87: In usage, I'm not sure if we should be asking developers to make the tint prop all-upper or all-lower.

Either works, but I'd want to keep it consistent across govuk-react. Off the top of my head, I can't think of anywhere we're accepting enums at the moment, but we should check. If this is the first instance we're setting a precedence and I'd want to make sure we're happy.

@stevesims @Loque- @marksy any preference?

@domdomegg am I right in assuming your preference would be all-upper for the prop value?

domdomegg: I'm not sure either.

I think my preference would actually be all-lower, but I don't really have a good reason for that (I think I've just used other libraries where it's been all lower and so it looks more 'normal to me').

Reason I went for all-upper here was copying the style of `Heading` which for its size prop states: "Visual size level, accepts: `XLARGE`, `LARGE`, `MEDIUM`, `SMALL`, `XL`, `L`, `M`, `S` or a numeric size that fits in the GDS font scale list".

But looking around other components, I think we also use all-lowercase e.g. for Grid Col's setWith: "Explicitly set column to width using value or descriptive string (`one-quarter`, `one-third`, `one-half`, `two-thirds`, `three-quarters`, `full`)"

components/tag/src/stories.tsx
```diff
@@ -13,11 +14,109 @@ export default {
...
 Default.args = {
   children: 'beta',
-  inactive: false,
 };
 
 export const Inactive: Story = (args) => <Tag {...args} />;
 Inactive.args = {
-  children: 'beta',
-  inactive: true,
+  children: 'inactive',
```
penx, line 21: this should have triggered a change in visual regression, but for some reason Chromatic didn't pick it up - I'll need to look in to this

components/tag/src/index.tsx
penx: yep size prop on heading is all caps:

https://github.com/govuk-react/govuk-react/blob/01164b69da02cb78ce770d44a479f73e93d84c9b/packages/constants/src/typography.tsx#L152-L159

https://github.com/govuk-react/govuk-react/blob/01164b69da02cb78ce770d44a479f73e93d84c9b/components/heading/src/heading.tsx#L24

https://github.com/govuk-react/govuk-react/blob/01164b69da02cb78ce770d44a479f73e93d84c9b/components/heading/src/stories.tsx#L41

And setWidth prop on GridCol is all-lower:

https://github.com/govuk-react/govuk-react/blob/01164b69da02cb78ce770d44a479f73e93d84c9b/components/grid-col/src/index.tsx#L55

https://github.com/govuk-react/govuk-react/blob/01164b69da02cb78ce770d44a479f73e93d84c9b/packages/constants/src/spacing.tsx#L83-L88

https://github.com/govuk-react/govuk-react/blob/01164b69da02cb78ce770d44a479f73e93d84c9b/components/grid-col/src/stories.tsx#L27


Ok as we're already inconsistent I'm happy to keep this as-is as we can look in to consistent naming of enum props in future, I'll create a ticket.

penx: #1099

components/tag/src/stories.tsx
```diff
@@ -13,11 +14,109 @@ export default {
...
+  tint: 'GREY',
```
penx, line 22: by deprecating `inactive` we are implying it will still work for now but will be removed in a future update.

Please can we keep the visual regression test for `inactive: true`. If you want, you could add a note to the story that this is deprecated:

```ts
Inactive.parameters = {
  docs: {
    storyDescription: 'deprecated',
  },
};
```

penx (changes requested):
Please retain the visual regression test for the inactive prop

components/tag/src/stories.tsx
```diff
@@ -13,11 +14,109 @@ export default {
...
+};
```
penx, line 23:
```suggestion
  children: 'inactive',
};
Inactive.parameters = {
  docs: {
    storyDescription: 'deprecated', 
    },
};
```

penx (approved):

components/tag/src/stories.tsx
```diff
@@ -13,11 +14,113 @@ export default {
...
 Default.args = {
   children: 'beta',
-  inactive: false,
 };
 
 export const Inactive: Story = (args) => <Tag {...args} />;
 Inactive.args = {
-  children: 'beta',
-  inactive: true,
+  children: 'inactive',
```
penx, line 21:
```suggestion
  children: 'inactive',
  inactive: true,
```

components/tag/src/stories.tsx
```diff
@@ -13,11 +14,114 @@ export default {
...
 
 export const Inactive: Story = (args) => <Tag {...args} />;
 Inactive.args = {
-  children: 'beta',
+  children: 'inactive',
   inactive: true,
 };
+Inactive.parameters = {
+  docs: {
+    storyDescription: 'deprecated',
```
penx, line 26:
```suggestion
    storyDescription: 'deprecated',
```

PR #1079, opened 2022-04-30
Support as= with strict TypeScript

domdomegg:
**Checklist**:

* [x] Documentation (N/A)
* [x] Tests
* [x] Ready to be merged

**Motivation**:

When consuming this library with TypeScript, where there are custom overriden `as` props they are often typed with `React.ComponentType`, which means really means `React.ComponentType<{}>`, i.e. a React component without any props.

When using TypeScript in non-strict mode, it allows you to pass in a `React.ComponentType<Something>` for `React.ComponentType<{}>`, even though really it shouldn't - that's why it's been working previously.

This PR corrects the type by expanding it to `React.ComponentType<any>`. While generally the use of `any` is discouraged, here it is actually the right choice given we're saying that our `as` can be a component that takes any props. This strategy is used by many well-known libraries, including react itself and react-router:
- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/310abbeba12679e25003079ebe9f011c2c95092c/types/react/index.d.ts#L828
- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/310abbeba12679e25003079ebe9f011c2c95092c/types/react-router/index.d.ts#L92

In addition to this, where we have a storybook with react-router-dom it didn't do any type-checking because we hadn't installed the types for react-router-dom as dev dependency. I've installed them so it does a bit more checking. Unfortunately enabling strict in tsconfig didn't work on this repo - think there may be other things to fix. So for now it's not as safe as it could be... future work I guess :)

For searchability, the TypeScript error you get with this library looks something like (when used with a Button and react-router-dom v6):

```
No overload matches this call.
  Overload 1 of 3, '(props: ButtonPropsWithoutAs, ref?: ButtonRefType | undefined): ReactElement<ButtonPropsWithoutAs, string | JSXElementConstructor<any>>', gave the following error.
    Type 'ForwardRefExoticComponent<LinkProps & RefAttributes<HTMLAnchorElement>>' is not assignable to type 'undefined'.
  Overload 2 of 3, '(props: ButtonPropsWithAs<string | ComponentType<{}>, string | ComponentType<{}>>, ref?: Ref<string | ComponentType<{}>> | undefined): ReactElement<...>', gave the following error.
    Type 'ForwardRefExoticComponent<LinkProps & RefAttributes<HTMLAnchorElement>>' is not assignable to type 'string | ComponentType<{}> | undefined'.
      Type 'ForwardRefExoticComponent<LinkProps & RefAttributes<HTMLAnchorElement>>' is not assignable to type 'FunctionComponent<{}>'.
        Types of parameters 'props' and 'props' are incompatible.
          Type '{}' is not assignable to type 'LinkProps & RefAttributes<HTMLAnchorElement>'.
            Property 'to' is missing in type '{}' but required in type 'LinkProps'.
  Overload 3 of 3, '(props: ButtonOwnProps): ReactElement<any, string | JSXElementConstructor<any>> | null', gave the following error.
    Type '{ children: string; as: ForwardRefExoticComponent<LinkProps & RefAttributes<HTMLAnchorElement>>; to: string; icon: Element; start: true; onClick: () => void; }' is not assignable to type 'IntrinsicAttributes & ButtonOwnProps'.
      Property 'as' does not exist on type 'IntrinsicAttributes & ButtonOwnProps'.ts(2769)
index.d.ts(55, 5): The expected type comes from property 'as' which is declared here on type 'IntrinsicAttributes & { form?: string | undefined; slot?: string | undefined; style?: CSSProperties | undefined; title?: string | undefined; ... 271 more ...; buttonTextColour?: string | undefined; } & { ...; } & { ...; }'
index.d.ts(53, 5): 'to' is declared here.
index.d.ts(59, 5): The expected type comes from property 'as' which is declared here on type 'IntrinsicAttributes & ButtonPropsWithAs<string | ComponentType<{}>, string | ComponentType<{}>>'
```


penx (approved):

penx:
@domdomegg thanks for this! Really appreciated.

Please could you resolve conflicts and fix linting issues?

domdomegg:
Hi @penx, thank you for the excellent library, I'm happy to contribute :)

I've resolved the conflicts and lint issues.

penx:
yep this seems to correlate with styled-components types too:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/ae3f89d525dca7f7073e61e74cee1871a018ea98/types/styled-components/index.d.ts#L91-L92

thanks!

penx:
I've created https://github.com/govuk-react/govuk-react/issues/1098 for enabling strict in tsconfig, agreed we should do this

Issue #1138, opened 2022-07-07
H1 tag type error


linusnorton:
**Describe the bug**
I get a typescript compilation error when using H1 tags:

```
TS2559: Type '{ children: string; }' has no properties in common with type 'IntrinsicAttributes & HeadingOwnProps & WithWhiteSpaceProps'
```

**To Reproduce**
```
import { H1 } from 'govuk-react';

<H1>Text</H1>
```

**Expected behavior**
No errors

**Screenshots**
N/A

penx:
AFAIK this is an issue with the React 18 typings removing `children` being there by default

https://stackoverflow.com/questions/71788254/react-18-typescript-children-fc

We'll need to address this in govuk-react by adding the children prop in manually for all components and test these against the React 18 types.

Would you like to submit a pull request for this?

linusnorton:
Resolved in 0.10.4

PR #1147, opened 2022-08-19
Add props for children to headers

linusnorton:
Fix for https://github.com/govuk-react/govuk-react/issues/1138

Some thoughts:

- I tested this by temporarily upgrading the React type defs to v18 and then reverted as the upgrade to React 18 seemed like a fairly big job
- I don't know what other components will also have a missing `children` prop. I've only fixed headers.
- I don't know what an appropriate comment for the property would be

**Checklist**:

* [ ] Documentation
* [ ] Tests
* [ ] Ready to be merged

### DIFF
--- a/components/heading/src/heading.tsx
+++ b/components/heading/src/heading.tsx
@@ -69,6 +69,7 @@ export const Heading: HeadingType = ({ level, size, ...props }: HeadingOwnProps 
 Heading.defaultProps = {
   level: undefined,
   size: 'XLARGE',
+  children: undefined,
 };
 
 Heading.displayName = 'Heading';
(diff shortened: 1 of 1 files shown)


penx:
Thanks for the PR!

> I tested this by temporarily upgrading the React type defs to v18 and then reverted as the upgrade to React 18 seemed like a fairly big job

Ok great, it would be awesome if we had a way to run `yarn test:types` against the React 18 types but I can't think of a way off the top of my head.


> I don't know what other components will also have a missing children prop. I've only fixed headers.

Running `yarn test:types` when using React types should show you any other components, as far as I can see it's only Headers and some storybook-only components that aren't published.

> I don't know what an appropriate comment for the property would be

I don't think we need a comment for the children prop on Headers

penx (approved):

covolution:
When is the next release with this fix in?

penx:
@covolution thanks for the reminder! I’ll try get a release out this week

penx:
Released in 0.10.4

PR #1186, opened 2023-01-23
Microsoft Edge fix for mismatch font weight

rottitime:
**Checklist**:

* [ ] Documentation
* [ ] Tests
* [x] Ready to be merged

### DIFF
--- a/components/top-nav/src/atoms/icon-title/index.tsx
+++ b/components/top-nav/src/atoms/icon-title/index.tsx
@@ -12,6 +12,7 @@ const IconContainer = styled('span')({
 });
 const TitleContainer = styled('span')({
   display: 'inline-block',
+  fontWeight: 700,
 });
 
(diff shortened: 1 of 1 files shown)


components/top-nav/src/index.tsx
```diff
@@ -84,7 +84,7 @@ export class TopNav extends Component<TopNavProps, { navigationOpen: boolean }>
 
   static displayName = 'TopNav';
 
-  static defaultProps = {
+  static defaultProps: TopNavProps = {
```
penx, line 87: thanks for the PR!

Any reason this was added in? I'm not sure defaultProps should be typed as TopNavProps.

If a required prop was added in the future, it wouldn't be required in the default props.

I think this is fine untyped, as it was.

rottitime: thanks for the feedback @penx I have reverted this now

penx (approved):

penx:
Thanks @rottitime!

Issue #597, opened 2019-03-11
Small Checkbox and Radio


stevesims:
For non-public-facing UK gov systems smaller sized input elements, especially checkbox and radio components, are often supported.  Examples of these can be found in the Home Office Design System:
https://home-office-digital-patterns.herokuapp.com/components/checkboxes
https://home-office-digital-patterns.herokuapp.com/components/radios

We should consider ways of adding support for such customisation.

There are potentially three options here:

1) Add a `small` prop that supports a (fixed) small size to these components
2) Add support for theming (see #578), and use that to support size variations
3) Have the atom (un-labelled checkbox/radio) exported from the component and used via a prop, allowing it to be re-styled

Or potentially some combination of all three options.

stevesims:
Home Office sass files with these alternate designs can be found here:
https://github.com/UKHomeOffice/design-system/tree/master/app/assets/sass/components

penx:
govuk-frontend now provides small radio options:

https://design-system.service.gov.uk/components/radios/small/index.html

rottitime:
I have developed an update for this . I have also added type support as noticed `Size` is something we use for many elements

Can someone please review my PR?
https://github.com/govuk-react/govuk-react/pull/1193

PR #1193, opened 2023-02-23
Feature/radio checkout size

rottitime:
Closes: https://github.com/govuk-react/govuk-react/issues/597
Closes: https://github.com/i-dot-ai/govuk-react/issues/1

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 # dependencies
 /node_modules
-package-lock.json
 
 # testing
 /coverage
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # dependencies
 /node_modules
+package-lock.json
 
 # testing
 /coverage
@@ -47,3 +48,4 @@ govuk-react-*.tgz
 !.yarn/plugins
 !.yarn/releases
 !.yarn/versions
+yarn.lock
--- a/.gitignore
+++ b/.gitignore
@@ -48,4 +48,3 @@ govuk-react-*.tgz
 !.yarn/plugins
 !.yarn/releases
 !.yarn/versions
-yarn.lock
--- a/components/checkbox/src/index.tsx
+++ b/components/checkbox/src/index.tsx
@@ -10,6 +10,7 @@ import HintText from '@govuk-react/hint-text';
 import { FOCUS_COLOUR } from 'govuk-colours';
 import { BORDER_WIDTH, BORDER_WIDTH_FORM_ELEMENT, FOCUS_WIDTH, SPACING_POINTS } from '@govuk-react/constants';
 import { spacing, typography } from '@govuk-react/lib';
+import type { Size } from '@govuk-react/lib';
 
 const checkboxSize = SPACING_POINTS[7];
 const checkboxSizeSmall = SPACING_POINTS[5];
@@ -23,7 +24,7 @@ const StyledCheckbox = styled('label')(
     minHeight: checkboxSize,
     padding: `0 0 0 ${checkboxSize}px`,
     clear: 'left',
-    '&.size-small': {
+    '&.size-s': {
       paddingRight: 22,
       span: {
         paddingLeft: 0,
@@ -119,7 +120,7 @@ const StyledCheckboxHint = styled(HintText)({
  */
 const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
   ({ children, className, hint, size, ...props }: CheckboxProps, ref) => (
-    <StyledCheckbox className={`${className} size-${size}`}>
+    <StyledCheckbox className={`${className} ${`size-${size}`.toLowerCase()}`}>
       <StyledInput type="checkbox" {...props} ref={ref} />
       <StyledLabel>{children}</StyledLabel>
       {hint && <StyledCheckboxHint>{hint}</StyledCheckboxHint>}
@@ -148,7 +149,7 @@ export interface CheckboxProps extends Omit<React.InputHTMLAttributes<HTMLInputE
   /**
    * Size of checkbox
    */
-  size?: 'medium' | 'small';
+  size?: `${Size.MEDIUM}` | `${Size.SMALL}`;
 }
 
 export default Checkbox;
--- a/components/checkbox/src/index.tsx
+++ b/components/checkbox/src/index.tsx
@@ -30,18 +30,18 @@ const StyledCheckbox = styled('label')<Pick<CheckboxProps, 'size'> & WithWhiteSp
       paddingRight: 22,
       span: {
         paddingLeft: 0,
-      },
-      'span::before': {
-        width: checkboxSizeSmall,
-        height: checkboxSizeSmall,
-        top: 8,
-      },
-      'span::after': {
-        top: 15,
-        left: 6,
-        width: 11,
-        height: 4.5,
-        borderWidth: '0 0 3px 3px',
+        ':before': {
+          width: checkboxSizeSmall,
+          height: checkboxSizeSmall,
+          top: 8,
+        },
+        ':after': {
+          top: 15,
+          left: 6,
+          width: 11,
+          height: 4.5,
+          borderWidth: '0 0 3px 3px',
+        },
       },
     },
   spacing.withWhiteSpace({ marginBottom: 2 })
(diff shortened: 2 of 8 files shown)


rottitime:
Hi, any updates on this? @penx

.gitignore
```diff
@@ -46,4 +47,5 @@ govuk-react-*.tgz
 !.yarn/patches
 !.yarn/plugins
 !.yarn/releases
-!.yarn/versions
\ No newline at end of file
+!.yarn/versions
+yarn.lock
```
penx, line 51: please can you revert the changes to this file?

components/checkbox/src/index.tsx
```diff
@@ -9,12 +9,14 @@ import styled from 'styled-components';
 import HintText from '@govuk-react/hint-text';
 import { FOCUS_COLOUR } from 'govuk-colours';
 import { BORDER_WIDTH, BORDER_WIDTH_FORM_ELEMENT, FOCUS_WIDTH, SPACING_POINTS } from '@govuk-react/constants';
-import { spacing, typography } from '@govuk-react/lib';
+import { spacing, typography, WithWhiteSpaceProps } from '@govuk-react/lib';
+import type { Size } from '@govuk-react/lib';
```
penx, line 13: What's the motivation for making this a generic abstraction called Size?

Are there any drawbacks to explicitly listing the possible values in each component that has a size prop?

penx: If it's an abstraction, is there a name we could use that is more specific to its use, such as `ComponentSizeVariant`?

components/checkbox/src/index.tsx
```diff
@@ -112,9 +133,10 @@ Checkbox.displayName = 'Checkbox';
 Checkbox.defaultProps = {
   hint: undefined,
   className: undefined,
+  size: 'medium',
 };
 
-export interface CheckboxProps extends React.InputHTMLAttributes<HTMLInputElement> {
+export interface CheckboxProps extends Omit<React.InputHTMLAttributes<HTMLInputElement>, 'size'> {
```
penx, line 139: Are there any downsides to preventing the user from accessing the html input size prop?

Would it be better to use a name for our size prop that didn't clash, such as `sizeVariant`?

components/checkbox/src/stories.tsx
```diff
@@ -18,10 +18,17 @@ Default.args = {
...
+CheckboxSmall.args = {
+  children: 'Small checkbox',
+  size: 'small',
+};
+
 export const CheckboxDisabled = Template.bind({});
 CheckboxDisabled.args = {
   children: 'Disabled checkbox option',
   disabled: true,
+  size: 100,
```
penx, line 31: I'm not sure how `100` is valid here?

components/radio/src/__snapshots__/test.tsx.snap
```diff
@@ -268,12 +271,15 @@ exports[`Radio disabled: disabled 1`] = `
...
   inline={false}
+  size="medium"
 >
   <src__Label
     inline={false}
+    size="medium"
   >
     <label
       className="c0"
+      size="medium"
```
penx, line 282: do we want the size prop to be sent to the label element?

penx:
Thanks for the PR @rottitime! Sorry I've been busy but have left some feedback and should have some time to look again tomorrow.

.gitignore
rottitime: thanks will do

components/checkbox/src/index.tsx
rottitime: @penx I noticed we use below in some components. As the sizes are shared it would be really good to start combining in one common place. It would also mean the sizes can be consistent. Is that ok?

I will update:
- name to be `ComponentSizeVariant`
- Values to match existing values: `XLARGE`, `LARGE`, `MEDIUM`, `SMALL`

```
`XLARGE`, `LARGE`, `MEDIUM`, `SMALL`, `XL`, `L`, `M`, `S`

File: components/checkbox/src/stories.tsx (hunk 4 above)
Comment (rottitime) [reply components/checkbox/src/stories.tsx]: Well spotted. This one was accidental

File: components/radio/src/__snapshots__/test.tsx.snap (hunk 5 above)
Comment (rottitime) [reply components/radio/src/__snapshots__/test.tsx.snap:282]: @penx the jest snapshots differ from what we have printed to screen. The real output is as:

```
<label class="sc-gsBqHK dEjzGa"><input type="radio" class="sc-aKXVD eEwhHG"><span class="sc-kmATbt cLEvoP">Small radio</span></label>

rottitime:
> Thanks for the PR @rottitime! Sorry I've been busy but have left some feedback and should have some time to look again tomorrow.

I made the updates and given feedback to others. Can you review again please? :)

.gitignore
```diff
@@ -1,5 +1,6 @@
 # dependencies
 /node_modules
+package-lock.json
```
penx, line 3: I see gitignore as containing files that, under normal usage, could be on disk but we would want to ignore.

package-lock.json is not something users should have under normal usage, and is a sign that something has gone wrong (i.e. they've used npm instead of yarn)

I would prefer this was not added to gitignore

penx:
This looks great, thanks for taking the time to submit.

One minor comment on the gitignore, otherwise I think I'm happy with this approach.

@stevesims @Loque- any thoughts?

components/checkbox/src/index.tsx
penx: yes, agreed it would be good to make sure the strings we pass in to the sizeVariant props across components are consistent, so I agree with the abstraction for that

.gitignore
rottitime: thanks I understand. This has been reverted back too

penx (approved):

penx:
Vertical centering of the small checkbox at 320px looks off. Is this correct?

Screenshot 2023-03-03 at 12 43 19

rottitime:
> Vertical centering of the small checkbox at 320px looks off. Is this correct?
> 
> Screenshot 2023-03-03 at 12 43 19

Thanks @penx I made the fontsize and spacing more consistent for both Radio and checkboxes with:
https://design-system.service.gov.uk/components/checkboxes/small/index.html

rottitime:
@penx is there any further action from my side? I am assuming not. Thanks again for all your feedback

components/checkbox/src/stories.tsx
```diff
@@ -18,6 +18,12 @@ Default.args = {
   children: 'Example',
 };
 
+export const CheckboxSmall = Template.bind({});
+CheckboxSmall.args = {
+  children: 'Small checkbox',
+  sizeVariant: 'SMALL',
```
penx, line 24: I think the prop value that is passed in should be lowercase, i.e. it should be usable as `<Checkbox sizeVariant="small" />`

rottitime: @penx I don't mind but this clashes with other parts of your code. All sizes so far have been uppercase.
Wouldn't it be great if we can retro fit the type for `sizeVariant` on all existing component. It will keep everything more manageable from a code point of view?

If you still need it changed to lowercase, I can still do so?

Existing doc:
Screenshot 2023-03-08 at 20 46 47

penx: Ah ok. Thanks. Let me have a think

rottitime: > Ah ok. Thanks. Let me have a think

@penx do you have an update. 

I think we should keep the size names consistent with the rest of your code. If we can start making it TypeSafe for now then we can change it much easier later on. I am happy to go with whatever you decide

penx: sorry just to let you know I haven't forgotten about this but have had a busy few weeks - will hopefully find some time tomorrow to review

penx: ok great, confirmed Heading and Caption currently have a size prop that accepts a number or an uppercase string in this set.

We should be consistent here, and if we make a change it should be done across the library.

We're breaking from convention here and using `sizeVariant` instead of `size` so as not to conflict with an existing html attribute.

So... all good I think!

Are you hoping to see this in a release soon?

rottitime: > Are you hoping to see this in a release soon?

there was a urgency but no so much anymore. so approve and merge whenever you can :)

PR #1650, opened 2024-04-29
Remove defaultProps to support React 18.3.0

richrace:
See this release from React: https://github.com/facebook/react/releases/tag/v18.3.0

* [ ] Documentation - N/A
* [x] Tests
* [x] Ready to be merged


richrace:
@penx, Could you have a look and see what you think?

I've removed all the `defaultProps` on the components.

components/skip-link/src/index.tsx
```diff
@@ -29,9 +30,43 @@ export const SkipLink = styled('a')(
...
+  as?: never | undefined;
+  forwardedAs?: never | undefined;
+};
+
+type SkipLinkPropsWithAs<
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  AsC extends string | React.ComponentType<any>,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  FAsC extends string | React.ComponentType<any> = AsC
+> = StyledComponentProps<AsC, never, SkipLinkOwnProps, never, FAsC> & {
```
penx, line 67: I think we can remove the `any` by doing something like

```suggestion
type SkipLinkPropsWithAs<
  Props,
  AsC extends string | React.ComponentType< Props >,
  FAsC extends string | React.ComponentType< Props > = AsC
> = StyledComponentProps<AsC, never, SkipLinkOwnProps, never, FAsC> & {
```

but I guess this was copied from elsewhere, this isn't high priority, and perhaps something that should be addressed across all places it's done in its own PR

components/tabs/src/atoms/title/index.tsx
```diff
@@ -1,16 +1,21 @@
...
 });
 
-TabsTitle.defaultProps = {
-  children: 'Contents',
+interface TabsTitleType extends React.HTMLAttributes<HTMLHeadingElement> {
+  children?: string | React.ReactNode;
+}
+
+const TabsTitle = ({ children = 'Contents', ...props }: TabsTitleType): JSX.Element => {
+  return <StyledTabsTitle {...props}>{children}</StyledTabsTitle>;
```
penx, line 18: I guess we're losing the ref here (and elswhere we're introducing a new wrapper component to a styled component), which could be a breaking change for some users.

I don't see a need for a ref here specifically so not too concerned - but is 'creating a wrapper component' the best way to set default props on a styled component now? Perhaps we could use `.attrs` instead? https://styled-components.com/docs/api#.attrs

penx: as far as I can see we are using withConfig and attrs elsewhere in govuk-react, I guess for this behaviour?

penx:
2 minor bits of feedback but all good otherwise:

- https://github.com/govuk-react/govuk-react/pull/1650#discussion_r1586358684
- https://github.com/govuk-react/govuk-react/pull/1650#discussion_r1586362862

Thanks for the contribution!

components/skip-link/src/index.tsx
richrace: Yes, it was copied. I think it's best addressed in it's own PR.

components/tabs/src/atoms/title/index.tsx
richrace: The main issue giving the default value to `children = 'Contents'`. I don't really about the `props` or `attrs`.

Should I `TabsTitlePropsWithAs` etc here as well?

penx: > TabsTitlePropsWithAs

Ah yeah that's what's required to use forwardRef isn't it, sorry it's been a while since I wrote this.

Probably safest to do that here so that users don't lose refs, which may be rare but would produce annoying, hard to identify bugs.

penx:
The fixed width font has changed in some stories on Chromatic, I suspect due to differences elsewhere or within Chromatic but worth checking against the main branch to be certain.

richrace:
> The fixed width font has changed in some stories on Chromatic, I suspect due to differences elsewhere or within Chromatic but worth checking against the main branch to be certain.

I'm not sure what I'm meant to be looking at. I've compared Chromatic to main (GitHub Pages) and local (my branch), and it seems consistent (apart from Chromatic not having the intros and args).

stevesims:
besides the font differences, Chromatic is also showing what looks like a margin difference on a radio button story.  this PR seems to have a significantly reduced margin

richrace:
Interesting, I have not changed any of the stylings. 😕

penx:
> besides the font differences, Chromatic is also showing what looks like a margin difference on a radio button story. this PR seems to have a significantly reduced margin

@stevesims I think you're referring to [this](https://www.chromatic.com/test?appId=5ad61aab7f00090020dfdba8&id=6632709499e5e8189c463a7b), which I think is also the fixed width thing, there seems to be a `{}` on this page, coming from something displaying in fixed width, and it has changed in height

@richrace If these stories look the same in main and this branch then I think we're good, likely due to a change with chromatic's rendering since we last approved something.

The thing to look out for is the font family and size being used for the fixed width parts of the stories above. Are you able to open the Chromatic links and see the visual diffs?

richrace:
The `{}` is the same, font family is `monospace` and using `pre` tag with `margin: 1em 0px`

Code is the same, `<code>` and the font family is `monospace`

Bold is the same, `<strong>` and font family is `"nta" ,Arial, sans-serif`

Italics are the same, `<em>` and font family is `"nta" ,Arial, sans-serif`

So it's likely due to:

> likely due to a change with chromatic's rendering since we last approved something

richrace:
Are you happy for me to Accept the visual changes that https://www.chromatic.com/build?appId=5ad61aab7f00090020dfdba8&number=2659 have thrown up?

penx:
> Are you happy for me to Accept the visual changes that https://www.chromatic.com/build?appId=5ad61aab7f00090020dfdba8&number=2659 have thrown up?

Sure if you have access to accept? Accept the ones that are due to a difference in rendering fixed width fonts, which I think is all of them

richrace:
That's all done. 

Just the cypress tests to run and it should be good to merge and be released.

stevesims:
> @stevesims I think you're referring to [this](https://www.chromatic.com/test?appId=5ad61aab7f00090020dfdba8&id=6632709499e5e8189c463a7b), which I think is also the fixed width thing, there seems to be a `{}` on this page, coming from something displaying in fixed width, and it has changed in height

yes - that was the thing.  if you look at the "1 up" view then there's a green line shown right at the bottom

having taken a closer look directly checking out the two respective storybooks for each branch being compared, I can see no difference - they are the same height

this therefore seems to be another odd chromatic rendering issue, like the minor font rendering differences the other stories are showing - so it feels safe to ignore. 😀

richrace:
On my fork, the Cypress tests run (and pass), but they don't seem to get run on the PR. What am I missing here?

richrace:
This should be all good to go.

penx:
@richrace thanks!

Do you want to attempt a release and see how far you get?

https://github.com/govuk-react/govuk-react/blob/main/CONTRIBUTING.md#releasing

PR #1683, opened 2024-05-24
Remove eslint/no-explicit-any and replace with proper types/interfaces

richrace:
* [ ] Documentation
* [ ] Tests
* [ ] Ready to be merged


components/tabs/src/atoms/tab/index.tsx
```diff
@@ -74,8 +74,10 @@ const Tab: TabType = ({ selected = false, ...props }: TabOwnProps) => (
 
 interface TabType extends React.FC<TabOwnProps> {
   (props: TabPropsWithoutAs): React.ReactElement<TabPropsWithoutAs>;
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  <AsC extends string | React.ComponentType<any> = 'a', FAsC extends string | React.ComponentType<any> = AsC>(
```
penx, line 78: I think the `any` here is so that, if you are using the `as` prop, you can pass any props to the component that will then be forwarded on to the `as` component

